PR #5, opened 2021-09-06
Providers optional TOFU auth wire clients, ephemeral clients gc

leif:
This is https://github.com/katzenpost/server/pull/151 migrated to the monorepo.


server/internal/provider/provider.go
```diff
@@ -213,14 +241,29 @@ func (p *provider) worker() {
 
 	ch := p.ch.Out()
 
+	// Here we optionally set this GC timer. If unset the
+	// channel remains nil and has no effect on the select
+	// statement below. If set then the timer will periodically
+	// write to the channel triggering our GC routine.
+	var gcEphemeralClientGCTickerChan <-chan time.Time
+	if p.glue.Config().Provider.EnableEphemeralClients {
+		ticker := time.NewTicker(epochtime.Period)
```
david415, line 250: Should we GC ephemeral clients every epoch? It skips clients which are currently connected.

mixmasala: i suspect clients are disconnecting on epoch transitions but am not sure if this is reliably occuring - it would be good to try this with some clients and see if they do stay connected across epochs.

mixmasala: i suspect clients are disconnecting on epoch transitions but am not sure if this is reliably occuring - it would be good to try this with some clients and see if they do stay connected across epochs.

PR #8, opened 2021-09-07
Replace loop with echo service

david415:
Major breaking change: include a built in echo service and get rid of the built in loop service. Why would we want to do this? If at any time we are designing a mixnet protocol that requires a "cross over point" then an echo service functions as such. The sender encodes the final destination in the Sphinx SURB it stuffs into the beginning of the Sphinx packet payload.

Probably this pull request should also modify minclient, client, catshadow, unit tests, docker and anything else that uses loop; it should replace loop with echo.

So far this pull-request only modifies the mix server. Work-in-progress.


mixmasala:
I found a few more references to loop and replaced them.

PR #10, opened 2021-09-09
Unfuck length prefixing

david415:
this isn't ready to be merged yet


david415:
This PR now works. This makes it so the client and server handle length prefixing instead of each client and server plugin having to do it.HOWEVER we should also make the mix server NOT crash if length prefixing is wrong.

david415:
I added one more correction:

don't make the mix server crash if length prefixing is incorrect

Instead print an error in the log and drop the packet

PR #18, opened 2021-09-12
Update memspool and panda server plugins

mixmasala:
This PR adds:
Updates memspool and panda services to use the new server plugin system
Makes catshadow clients send/receive cbor encoded messages
Removes length prefixed encoding where cbor already handles length prefixed decoding internally.


Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM golang:alpine AS builder
```
david415, line 1: why are you reverting the docker file back to an ancient version?

david415 (changes requested):
merge it but first fix it so it doesn't merge that old obsolete docker file.

mixmasala:
this PR now includes changes to panda service as well as the memspool service. the catshadow docker-based integration tests are now passing.

Dockerfile
mixmasala: This Dockerfile was removed on main and I guess I didn't remove it on this branch when merging main into it... Removed now.

mixmasala:
I also removed the command padding from memspool. I'm not sure why this was padding the messages here as the sphinx payloads are also padded.

mixmasala (commented):
we still need to update the reunion service as well but that makes use of a feature that lets plugins publish a set of parameters in the mixdescriptor.

we dont have a way to fetch plugin parameters with the new plugin system, though.

PR #21, opened 2021-09-16
Add reunion commandio

mixmasala:
this PR is to track work on updating the reunion service


mixmasala:
this is still broken, and the implementation does not match https://codeberg.org/rendezvous/reunion

PR #25, opened 2021-11-02
eddsa/blinded25519: the Tor V3 onion key blinding scheme for ed25519

threebithacker:
This is a port of the scheme used by Tor. It will be useful as a building block
for features that are supposed to be unlinkable-to-some-but-not-everybody, where
we can use it to derive spool IDs and other identifiers asynchronously.

### DIFF
--- a/core/crypto/eddsa/blinded25519.go
+++ b/core/crypto/eddsa/blinded25519.go
@@ -1,0 +1,169 @@
+// blinded25519.go - blinded EdDSA signatures
+// License: AGPL version 3
+// Copyright: 2021 - Anonymous contributor
+
+// This module implements the blinded signature scheme as used in
+// Tor for OnionV3, see A.2. Tor's key derivation scheme:
+// https://github.com/torproject/torspec/blob/main/rend-spec-v3.txt#L2268-L2326
+// Note that this implementation of the scheme uses different constants
+// for hashing the (factor)s and as thus is not wire format compatible.
+
+// Note that the reason that Tor doesn't need a separate Sign() function
+// for blinded private keys is that they use a different in-memory
+// representation of secret keys where the nonce seed is stored in expanded
+// form (instead of deriving it with sha512 for each signature)
+
+// It would be kind of useful to have
+// BlindedPrivateKey.FromBytes()
+// BlindedPrivateKey.ToBytes()
+// to enable passing around the derived keys, for instance to pass on
+// delegated privileges. Unfortunately checking the private scalars does
+// not seem trivial (since they are "malformed", ie no clamping), so it
+// seems like it could be a great footgun. Leaving it out for now, but
+// it's something to consider for the future I guess.
+
+// Note that use of Scalar.SetUniformBytes() is a bit unorthodox:
+// The API was designed to take 64 uniformly random bytes,
+// which are then reduced mod L. We "abuse" it by providing
+// our 32byte scalars as the 32 lower bytes and the upper 32 bytes all-zero
+// --- This way we get to initialize Scalar objects without the usual clamping.
+// There doesn't seem to be another API that lets us initialize 32-byte scalars
+// directly.
+
+// And lastly, note that it would be interesting to come up with a
+// blinding scheme that preserves the "validity" of the derived secret key
+// for use with x25519, perhaps a way to modify `factor` such that
+//            clamp(a)*clamp(factor)
+// ===  clamp(clamp(a)*clamp(factor))
+
+package eddsa
+
+import (
+	"crypto/ed25519"
+	"crypto/sha512"
+	"filippo.io/edwards25519"
+)
+
+type BlindedPrivateKey struct {
+	blinded ed25519.PrivateKey
+}
+
+func (bpk *BlindedPrivateKey) PublicKey() *PublicKey {
+	pub := new(PublicKey)
+	err := pub.FromBytes(bpk.blinded[32:])
+	if err != nil {
+		// This should not be possible:
+		panic(err)
+	}
+	return pub
+}
+
+func (privateKey *BlindedPrivateKey) Sign(message []byte) [64]byte {
+	signature := [64]byte{}
+	// vendored version of ed25519.sign() except it uses the
+	// secret/private scalar directly as expandedSecretKey
+	// instead of deriving from hash each time.
+	// It is a bit unfortunate that this needs to be here...
+	blindedSecretKey := [64]byte{}
+	copy(blindedSecretKey[:], privateKey.blinded[:32])
+	expandedSecretKey := new(edwards25519.Scalar)
+	expandedSecretKey.SetUniformBytes(blindedSecretKey[:])
+
+	// secret_salt <- sha512(private)
+	digest1 := sha512.Sum512(privateKey.blinded[:32])
+
+	// very important that nonce is never repeated for different messages;
+	// ensured here by deriving nonce from a hash of the message.
+	// the second appendix of secret_salt below is not part of the spec:
+	// nonce := sha512(secret_salt[32:64] || message || secret_salt[33:64])
+	h := sha512.New()
+	h.Write(digest1[32:])
+	h.Write(message)
+	h.Write(digest1[33:])
+	messageDigest := h.Sum(nil)
+
+	// messageDigestReduced <- (nonce mod L)
+	mdReduced, _ := new(edwards25519.Scalar).SetUniformBytes(messageDigest[:64])
+
+	// R <- messageDigestReduced * B
+	var encodedR [32]byte
+	copy(encodedR[:], new(edwards25519.Point).ScalarBaseMult(mdReduced).Bytes())
+
+	// hramDigestReduced := sha512(R || public || message) mod L
+	h.Reset()
+	h.Write(encodedR[:])
+	h.Write(privateKey.blinded[32:])
+	h.Write(message)
+	hramDigest := h.Sum(nil)
+	hramDigestReduced, _ := new(edwards25519.Scalar).SetUniformBytes(hramDigest[:])
+	//     s := (hRamDigestReduced*secret + messageDigestReduced) mod L
+	// <-> s := (   (sha512(R || public || message) mod L)*secret
+	//            + (nonce mod L)
+	//          ) mod L
+	s_new := new(edwards25519.Scalar).MultiplyAdd(hramDigestReduced,
+		expandedSecretKey,
+		mdReduced)
+
+	copy(signature[:], encodedR[:])
+	copy(signature[32:], s_new.Bytes())
+	return signature
+}
+
+func hash_factor_inplace(factor *[32]byte) {
+	h := sha512.Sum512_256(factor[:])
+	copy(factor[:], h[:])
+	// needs to be clamped for scalar multiplication,
+	// but Scalar.SetBytesWithClamping takes care of that.
+}
+
+func (secret *PrivateKey) Blind(factor [32]byte) *BlindedPrivateKey {
+	hash_factor_inplace(&factor)
+	factor_sc, err := new(edwards25519.Scalar).SetBytesWithClamping(factor[:])
+	if err != nil {
+		// This only happens if factor is not 32 bytes.
+		panic(err)
+	}
+
+	// Here digest is the actual secret key derived from the seed:
+	digest := sha512.Sum512(secret.Bytes()[:32])
+	bb_sc, err := new(edwards25519.Scalar).SetBytesWithClamping(digest[:32])
+
+	// (ab + c) mod l = ScMulAdd(out, a, b, c)
+	// a := factor
+	// b := secret
+	// c := 0
+	// the (c*B) multiplication here is unfortunate but that was the easiest API
+	// to use in edwards25519, so...
+	// oo <- factor * bb + 0*B
+	oo_sc := new(edwards25519.Scalar).Multiply(factor_sc, bb_sc)
+
+	newsec := make(ed25519.PrivateKey, 64)
+	copy(newsec[:32], oo_sc.Bytes())
+
+	// calculates and sets the public key corresponding to the
+	// private scalar (used when we have derived a new private scalar)
+	newsec_a := [64]byte{}
+	copy(newsec_a[:], oo_sc.Bytes()[:32])
+	newsec_scalar, _ := new(edwards25519.Scalar).SetUniformBytes(newsec_a[:])
+	pkxA_n := new(edwards25519.Point).ScalarBaseMult(newsec_scalar)
+	copy(newsec[32:], pkxA_n.Bytes())
+
+	bpk := new(BlindedPrivateKey)
+	bpk.blinded = newsec
+	return bpk
+}
+
+func (mypub *PublicKey) Blind(factor [32]byte) *PublicKey {
+	// out <- factor*pkA + zero*Basepoint
+
+	hash_factor_inplace(&factor)
+	factor_sc, _ := new(edwards25519.Scalar).SetBytesWithClamping(factor[:])
+	out, _ := new(edwards25519.Point).SetBytes(mypub.Bytes())
+	newkey := new(PublicKey)
+	err := newkey.FromBytes(out.ScalarMult(factor_sc, out).Bytes())
+	if err != nil {
+		// Again this should not happen; but in case it does:
+		panic(err)
+	}
+	return newkey
+}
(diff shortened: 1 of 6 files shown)


core/crypto/eddsa/eddsa.go
```diff
@@ -144,13 +145,9 @@ func (k *PublicKey) ToPEMFile(f string) error {
 
 // ToECDH converts the PublicKey to the corresponding ecdh.PublicKey.
 func (k *PublicKey) ToECDH() *ecdh.PublicKey {
-	var dhBytes, dsaBytes [32]byte
-	copy(dsaBytes[:], k.Bytes())
-	defer utils.ExplicitBzero(dsaBytes[:])
-	extra25519.PublicKeyToCurve25519(&dhBytes, &dsaBytes)
-	defer utils.ExplicitBzero(dhBytes[:])
+	ed_pub, _ := new(edwards25519.Point).SetBytes(k.Bytes())
```
mixmasala, line 148: is there an equivalent to ExplicitBzero for edwards25519.Point ? I looked at the available methods but did not see a way to clear this before letting it be garbage collected

core/crypto/eddsa/blinded25519.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func hash_factor_inplace(factor *[32]byte) {
+	h := sha512.Sum512_256(factor[:])
+	copy(factor[:], h[:])
+	// needs to be clamped for scalar multiplication,
+	// but Scalar.SetBytesWithClamping takes care of that.
+}
+
+func (secret *PrivateKey) Blind(factor [32]byte) *BlindedPrivateKey {
```
mixmasala, line 119: it would be better to pass a reference to factor or use the []byte type

core/crypto/eddsa/blinded25519.go
```diff
@@ -0,0 +1,169 @@
...
+	// to use in edwards25519, so...
+	// oo <- factor * bb + 0*B
+	oo_sc := new(edwards25519.Scalar).Multiply(factor_sc, bb_sc)
+
+	newsec := make(ed25519.PrivateKey, 64)
+	copy(newsec[:32], oo_sc.Bytes())
+
+	// calculates and sets the public key corresponding to the
+	// private scalar (used when we have derived a new private scalar)
+	newsec_a := [64]byte{}
```
mixmasala, line 145: https://github.com/FiloSottile/edwards25519/blob/main/scalar.go#L93
already does a copy so this is not necessary

ie just do
```newsec_scalar, _ := new(edwards25519.Scalar).SetUniformBytes(oo_src.Bytes()[:32])```

File: core/crypto/eddsa/blinded25519.go [hunk 4]
```diff
@@ -0,0 +1,169 @@
...
+	newsec_scalar, _ := new(edwards25519.Scalar).SetUniformBytes(newsec_a[:])
+	pkxA_n := new(edwards25519.Point).ScalarBaseMult(newsec_scalar)
+	copy(newsec[32:], pkxA_n.Bytes())
+
+	bpk := new(BlindedPrivateKey)
+	bpk.blinded = newsec
+	return bpk
+}
+
+func (mypub *PublicKey) Blind(factor [32]byte) *PublicKey {
```
mixmasala, line 156: i think it is better to pass factor by reference or slice type (and check length)

core/crypto/eddsa/blinded25519_test.go
```diff
@@ -0,0 +1,133 @@
...
+	// here we do scalar multiplication with L as the scalar;
+	// the result should be 1 if the public key is valid.
+	var result [32]byte
+	order_64 := [64]byte{0xed, 0xd3, 0xf5, 0x5c, 0x1a, 0x63, 0x12, 0x58, 0xd6, 0x9c,
+		0xf7, 0xa2, 0xde, 0xf9, 0xde, 0x14, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10}
+	order_sc, _ := new(edwards25519.Scalar).SetUniformBytes(order_64[:])
+	pkA_n, _ := new(edwards25519.Point).SetBytes(pk.Bytes())
+	pkA_n.ScalarMult(order_sc, pkA_n)
+	copy(result[:], pkA_n.Bytes())
```
mixmasala, line 25: why is this copy necessary for the comparison?

core/crypto/eddsa/blinded25519.go
```diff
@@ -0,0 +1,169 @@
...
+	pub := new(PublicKey)
+	err := pub.FromBytes(bpk.blinded[32:])
+	if err != nil {
+		// This should not be possible:
+		panic(err)
+	}
+	return pub
+}
+
+func (privateKey *BlindedPrivateKey) Sign(message []byte) [64]byte {
```
mixmasala, line 61: this would be better if it implemented core/crypto/cert.Signer:
```
// Signer signs messages.
type Signer interface {
        // Sign signs the message and returns the signature.
        Sign(msg []byte) []byte

        // Identity returns the Signer identity.
        Identity() []byte

        // KeyType returns the key type string.
        KeyType() string
}

```

core/crypto/eddsa/blinded25519.go
```diff
@@ -0,0 +1,169 @@
...
+	h.Write(message)
+	h.Write(digest1[33:])
+	messageDigest := h.Sum(nil)
+
+	// messageDigestReduced <- (nonce mod L)
+	mdReduced, _ := new(edwards25519.Scalar).SetUniformBytes(messageDigest[:64])
+
+	// R <- messageDigestReduced * B
+	var encodedR [32]byte
+	copy(encodedR[:], new(edwards25519.Point).ScalarBaseMult(mdReduced).Bytes())
```
mixmasala, line 90: it isn't necessary to copy encodedR twice - prefer:
```encodedR := new(edwards25519.Point).ScalarBaseMult(mdReduced).Bytes())```

File: core/crypto/eddsa/blinded25519.go [hunk 8]
```diff
@@ -0,0 +1,169 @@
...
+	hramDigestReduced, _ := new(edwards25519.Scalar).SetUniformBytes(hramDigest[:])
+	//     s := (hRamDigestReduced*secret + messageDigestReduced) mod L
+	// <-> s := (   (sha512(R || public || message) mod L)*secret
+	//            + (nonce mod L)
+	//          ) mod L
+	s_new := new(edwards25519.Scalar).MultiplyAdd(hramDigestReduced,
+		expandedSecretKey,
+		mdReduced)
+
+	copy(signature[:], encodedR[:])
```
mixmasala, line 107: see comment above re: encodedRk - this is then
``` copy(signature[:], encodedR)```

File: core/crypto/eddsa/blinded25519.go [hunk 9]
```diff
@@ -0,0 +1,169 @@
...
+func (privateKey *BlindedPrivateKey) Sign(message []byte) [64]byte {
+	signature := [64]byte{}
+	// vendored version of ed25519.sign() except it uses the
+	// secret/private scalar directly as expandedSecretKey
+	// instead of deriving from hash each time.
+	// It is a bit unfortunate that this needs to be here...
+	blindedSecretKey := [64]byte{}
+	copy(blindedSecretKey[:], privateKey.blinded[:32])
+	expandedSecretKey := new(edwards25519.Scalar)
+	expandedSecretKey.SetUniformBytes(blindedSecretKey[:])
```
mixmasala, line 70: it isn't necessary to copy into blindedSecretKey since SetUniformBytes does its own copy internally:, this can be replaced with
```
expandedSecretKey := new(edwards25519.Scalar)
expandedSecretKey.SetUniformBytes(privateKey.blinded[:32])
```

core/crypto/eddsa/blinded25519.go
```diff
@@ -0,0 +1,169 @@
...
+	// (ab + c) mod l = ScMulAdd(out, a, b, c)
+	// a := factor
+	// b := secret
+	// c := 0
+	// the (c*B) multiplication here is unfortunate but that was the easiest API
+	// to use in edwards25519, so...
+	// oo <- factor * bb + 0*B
+	oo_sc := new(edwards25519.Scalar).Multiply(factor_sc, bb_sc)
+
+	newsec := make(ed25519.PrivateKey, 64)
```
mixmasala, line 140: i think you want to use new(ed25519.PrivateKey)?

core/crypto/eddsa/blinded25519.go
```diff
@@ -0,0 +1,169 @@
...
+	copy(newsec[:32], oo_sc.Bytes())
```
mixmasala, line 141: and FromBytes

mixmasala (changes requested):
noted some extra copy() operations which can be omitted

mixmasala:
https://github.com/katzenpost/katzenpost/commit/249df24db7cf4eb4974336361ff96c18b9bf680b

(but somehow all the tests are broken o_O)

core/crypto/eddsa/blinded25519.go
```diff
@@ -0,0 +1,169 @@
...
+
+package eddsa
+
+import (
+	"crypto/ed25519"
+	"crypto/sha512"
+	"filippo.io/edwards25519"
+)
+
+type BlindedPrivateKey struct {
```
mixmasala, line 47: https://golang.org/doc/effective_go#commentary
"Inside a package, any comment immediately preceding a top-level declaration serves as a doc comment for that declaration. Every exported (capitalized) name in a program should have a doc comment."

core/crypto/eddsa/blinded25519.go
mixmasala: is the intent to modify the callers copy of factor? I refactored this a little bit and basically Blind(factor[32]byte) is passing a COPY of factor to Blind() and then modifying it. but the caller of Blind()'s copy of factor is not modified.

in my refactored version Blind() is instead passed a byte slice []byte and doesn't use hash_factor_inplace because that then DOES modify the caller of Blind()'s copy of factor (which caused the unit tests to fail, so I think that isn't the expected behavior)

mixmasala (changes requested):
updated some comments

core/crypto/eddsa/blinded25519.go
threebithacker: Yeah, this is to work around the `SetUniformBytes` thing as explained in the top of the file.

core/crypto/eddsa/blinded25519.go
threebithacker: What's the benefit of doing it that way?

core/crypto/eddsa/eddsa.go
david415: i'm guessing we'd have to fork the edswards25519 module to get it to wipe the bytes... i'm fine if you want to do that on a second pass. or maybe ask for the feature addition or submit a pull/merge request.

core/crypto/eddsa/blinded25519_test.go
david415: in general we don't use == for byte comparisons because it only works on arrays AND in this particular case pkA_n.Bytes() returns a byte slice. Therefore it would be more appropriate to just compare the byte slices directly with bytes.Equal (from the go std libarry "bytes" module)... and in a production module (this one is a test), we'd use hmac.Equal or the "subtle" module for constant time comparison.

david415: so to fix it... 

1. remove the var result declaration
2. remove the copy
3. perform comparison with bytes.Equal and return the bool

threebithacker:
> updated some comments

The intent is to take a copy of `factor`, not to modify the caller's copy. The caller might want to use the factor for something else, for instance when the caller wants to blind using the current epoch random value, which is the main intended use case for this function, so I think the code (and test) is correct.

core/crypto/eddsa/eddsa.go
```diff
@@ -238,16 +235,12 @@ func (k *PrivateKey) KeyType() string {
 
 // ToECDH converts the PrivateKey to the corresponding ecdh.PrivateKey.
 func (k *PrivateKey) ToECDH() *ecdh.PrivateKey {
-	var dsaBytes [64]byte
-	defer utils.ExplicitBzero(dsaBytes[:])
-	copy(dsaBytes[:], k.Bytes())
-
-	var dhBytes [32]byte
-	extra25519.PrivateKeyToCurve25519(&dhBytes, &dsaBytes)
-	defer utils.ExplicitBzero(dhBytes[:])
```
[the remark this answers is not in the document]
threebithacker: Oh, sorry, I did get rid of the `extra25519` ref, but `utils.ExplicitBZero` could stay.
I didn't / don't see the point of randomly littering the code with `explicit_bzero` calls; launching the application with `GODEBUG=clobberfree` as documented [here](https://pkg.go.dev/runtime#hdr-Environment_Variables) achieves the same with less room for mistakes.
Another approach would be to attach [runtime.SetFinalizer](https://pkg.go.dev/runtime#SetFinalizer) to various constructors throughout the codebase.

Again, I think it's a pointless to try to clear these values from freed memory, but I think spraying `explicit_bzero` calls by hand is an error-prone strategy to achieve that, and I think it makes the code harder to read.

core/crypto/eddsa/blinded25519.go
threebithacker: answered below re: copying factor.
Hmm, if we use a `[]byte` slice then we open up the possibility for API misuse passing a non-32-byte factor, and then we need a check for that and a corresponding exception/error, and then callers need to handle that. I think that adds unnecessary complications for the users compared to the cost of copying the additional eight bytes.

core/crypto/eddsa/blinded25519.go
threebithacker: For some reason this proposed change segfaults `go test`.
I don't understand why, since I agree with your interpretation that it should copy it.
Maybe we are using an older version of the library?

core/crypto/eddsa/blinded25519.go
threebithacker: not understood?

core/crypto/eddsa/blinded25519_test.go
threebithacker: I think I was getting confused by slice/byte stuff in Golang when I wrote this, I've updated with @david415's suggestion

core/crypto/eddsa/eddsa.go
threebithacker: marking resolved, we can discuss this in https://github.com/katzenpost/katzenpost/pull/25#discussion_r885850820 or in a separate issue thread

core/crypto/eddsa/eddsa.go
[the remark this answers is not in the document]
david415: @mixmasala the call to ExplicitBzero is no longer needed because of the extra25519 code removal. @threebithacker tried to explain this with a joke.

core/crypto/eddsa/blinded25519.go
david415: @threebithacker Please make Sign method return a byte slice []byte as per @mixmasala request... as it might be better if it complied with that Signer interface he pasted.

core/crypto/eddsa/blinded25519.go
david415: @mixmasala in general we don't want to modify blinding factors... instead we use blinding factors to "bind" a public key... mutate public key or return a new mutated public key.

core/crypto/eddsa/blinded25519.go
david415:
```
newsec := make(ed25519.PrivateKey, 64)
```
looks wrong. when we make types we should use a constructor most of the time.

core/crypto/eddsa/blinded25519.go
david415: @threebithacker he's suggesting that if you want to load raw bytes into a key that you should use the FromBytes method to do that.

david415:
We don't want to use ToECDH anywhere. All the places that's used are silly little conveniences where we have an eddsa identity key but we really also want a link key. In order to make the core/wire module use a generalized KEM I need to get rid of the ToECDH usage so that we can replace our current usage of ECDH wire protocol keys with a hybrid KEM wrapper type. Hope that makes sense and sorry for all the confusion. Next time you want to optimize some of our crypto talk with us and let's figure out if we actually want and need the optimization.

david415:
merged via another branch

PR #30, opened 2022-08-24
NIKE Sphinx, CTIDH and ECDH

david415:
This pull request teaches our sphinx module to use my new NIKE interface and NIKE interface implementations. These code changes will not result in any user visible feature additions.

Had to change the crypto/ecdh module very minimally to support the Nike interface... and therefore we can't actually merge this yet until the following tasks are completed:

TODO:
-----
1. Unfuck all the things that broke because of ecdh changes
2. Unfuck all the things that broke because of sphinx changes.
3. Make the server + minclient work with the new sphinx module without any feature additions, just make it use ecdh sphinx.

I plan to merge this once those tasks are done. Actual feature additions to the mix server and minclient will happen in future pull requests.


david415:
Oh um also this changeset includes an upgrade to our core/crypto/ecdh module such that it no longer uses the deprecated call to ScalarMult but instead uses X25519, see the docs:

https://pkg.go.dev/golang.org/x/crypto/curve25519#ScalarMult

david415:
continuing to work on this here --> https://github.com/katzenpost/katzenpost/pull/37

Issue #38, opened 2022-09-10
Replace our eddsa with a hybrid pq + classical signature scheme?


david415:
This signature scheme is at the root of the security dependency graph of the katzenpost mix network.
I made a hybrid Dilithium2-aes + Ed25519 work... however now I'm thinking it's better to use Sphincs+. So we're going to use a hybrid scheme with Sphincs+ and Ed25519... although one could argue Sphincs+ is all we need and the Ed25519 is unnecessary.

The easiest way to get Sphincs+ working with katzenpost is to take this existing pure Go implementation and make it serialize it's types properly so that we can actually use it. Currently I'm advising the author and he seems to be working on it:

https://github.com/kasperdi/SPHINCSPLUS-golang/issues/2

Otherwise we fork it and fix it ourselves OR write CGO bindings for the C reference implementation of Sphincs+, which is here: https://github.com/sphincs/sphincsplus

Later on we may want to write cgo bindings for the C reference implementation anyways for performance reasons... to make it go faster, perhaps using AES-NI hardware acceleration.

The code is already written. All the tests and integration tests pass. All that is left to do is make the most suitably paranoid hybrid PQ + classical signature scheme work with the interface i've create thus making it a one line code change.

@Yawning What do you think about this?

david415:
pull request work in progress here: https://github.com/katzenpost/katzenpost/pull/45

Yawning:
I'm not a huge fan of cgo, in that if I need to use AES-NI, I'll just write the assembly.  While I am still rather paranoid, I'm not sure PQ auth makes sense at the moment, because exploits for that require active adversaries with quantum computers.

PQ HFS matters because adversaries in the future can break forward-secrecy.  That sort of concern is (IMO) moot for authentication...

PR #45, opened 2022-09-16
Add signature interface

david415:
work-in-progress on this ticket here https://github.com/katzenpost/katzenpost/issues/38


david415:
i'm going to start over... because my changes to the authority weren't very good.

PR #46, opened 2022-09-24
hybrid pq signatures for katzenpost

david415:
fixed voting authority unit tests. New PQ + classical hybrid signature scheme for Katzenpost. That is to say, all mixnet components, mixes, providers and authority servers will now sign mix descriptors and PKI consensus documents using the new PQ hybrid signature scheme. With this change the PQNoise changes to the wire protocol will actually be meaningful.

https://github.com/katzenpost/katzenpost/issues/38

Currently it uses a simple hybrid signature scheme with Sphincs+ & ED25519.


authority/voting/client/client.go
```diff
@@ -211,14 +220,14 @@ func (p *connector) roundTrip(s *wire.Session, cmd commands.Command) (commands.C
...
-func (p *connector) allPeersRoundTrip(ctx context.Context, linkKey wire.PrivateKey, signingKey *eddsa.PublicKey, cmd commands.Command) ([]commands.Command, error) {
+func (p *connector) allPeersRoundTrip(ctx context.Context, linkKey wire.PrivateKey, signingKey sign.PublicKey, cmd commands.Command) ([]commands.Command, error) {
 	doneCh := make(chan interface{})
 	defer close(doneCh)
 	responses := []commands.Command{}
 	for _, peer := range p.cfg.Authorities {
 		conn, err := p.initSession(ctx, doneCh, linkKey, signingKey, peer)
 		if err != nil {
-			p.log.Noticef("pki/voting/client: failure to connect to Authority peer %s", peer.IdentityPublicKey)
+			p.log.Noticef("pki/voting/client: failure to connect to Authority peer %s", peer.IdentityPublicKeyPem)
```
mixmasala, line 230: Use the b64 encoded hash of the public identity key

authority/voting/client/client.go
```diff
@@ -154,23 +157,29 @@ func (p *connector) initSession(ctx context.Context, doneCh <-chan interface{},
 		}
 	}()
 
-	var ad []byte
-	if signingKey != nil {
-		ad = signingKey.Bytes()
+	peerLinkPublicKey, err := wire.DefaultScheme.PublicKeyFromPemFile(filepath.Join(p.cfg.DataDir, peer.LinkPublicKeyPem))
```
mixmasala, line 160: these changes break katzen, 
which serializes its configuration into the encrypted statefile
It would be better to have the right types in the Config that can be serialized to CBOR
and that way the client can also be flexible in where it reads the PKI root keys from (i.e. they may be embedded in a binary or stored in another format)

authority/voting/client/client.go
```diff
@@ -386,7 +395,12 @@ func New(cfg *Config) (pki.Client, error) {
 	c.pool = newConnector(cfg)
 	c.verifiers = make([]cert.Verifier, len(c.cfg.Authorities))
 	for i, auth := range c.cfg.Authorities {
-		c.verifiers[i] = cert.Verifier(auth.IdentityPublicKey)
+		_, authIdPublicKey := cert.Scheme.NewKeypair()
+		err := pem.FromFile(filepath.Join(c.cfg.DataDir, auth.IdentityPublicKeyPem), authIdPublicKey)
```
mixmasala, line 399: see above, I prefer to have the final types in Config, not paths to PEM files

authority/voting/server/server.go
```diff
@@ -200,24 +204,69 @@ func New(cfg *config.Config) (*Server, error) {
...
 	linkPrivateKeyFile := filepath.Join(s.cfg.Authority.DataDir, "link.private.pem")
 	linkPublicKeyFile := filepath.Join(s.cfg.Authority.DataDir, "link.public.pem")
-	if s.linkKey, err = scheme.Load(linkPrivateKeyFile, linkPublicKeyFile, rand.Reader); err != nil {
-		s.log.Errorf("Failed to initialize link key: %v", err)
-		return nil, err
+
+	var linkPrivateKey wire.PrivateKey = nil
+	var linkPublicKey wire.PublicKey = nil
+
+	if pem.BothExists(linkPrivateKeyFile, linkPublicKeyFile) {
```
mixmasala, line 241: I prefer to not have keyfiles external to the configuration file, and do not feel that the serialized publickeys are large enough to warrant this change. Can we just use the Marshaller interface to serialize PEM encoded keys from the Config fields?

mixmasala (changes requested):
This looks very promising and correct as far as I see - however, I would ask that we not store key material in external PEM files, but store the PEM encoded keys in the voting Client Config by using appropriate types that implement the Marshaler interfaces so that serialization to/from toml works and the PKI root can be flexibly loaded rather than requiring it be on the filesystem in PEM format.

authority/voting/client/client.go
david415: this latest change will print the key in PEM format which I like better because it says the key type

authority/voting/server/state.go
```diff
@@ -407,11 +428,9 @@ func (s *SharedRandom) Reveal() []byte {
 }
 
 func (s *state) reveal(epoch uint64) []byte {
-	if reveal, ok := s.reveals[epoch][s.identityPubKey()]; ok {
+	if reveal, ok := s.reveals[epoch][s.identityPubKeyHash()]; ok {
```
threebithacker, line 431: wait why don't we just do this when we populate `s.reveals[epoch][s.identityPubKeyhash()] = srv.Reveal()` ?
Anyway, patch here looks fine.

authority/voting/server/state.go
```diff
@@ -647,7 +668,7 @@ func (s *state) sendVoteToPeer(peer *config.AuthorityPeer, vote []byte, epoch ui
 // for our link layer wire protocol as specified by
 // the PeerAuthenticator interface in core/wire/session.go
 func (s *state) IsPeerValid(creds *wire.PeerCredentials) bool {
-	var ad [eddsa.PublicKeySize]byte
+	var ad [publicKeyHashSize]byte
 	copy(ad[:], creds.AdditionalData)
```
threebithacker, line 672: does this crash if there's more bytes in `creds.AdditionalData` than `publicKeyHashSize` ?

authority/voting/server/state.go
```diff
@@ -1393,14 +1424,17 @@ func (s *state) restorePersistence() error {
...
+				eDescsBkt := descsBkt.Bucket(epochBytes)
 				if eDescsBkt == nil {
 					s.log.Debugf("No persisted Descriptors for epoch: %v.", epoch)
 					continue
 				}
 
 				c := eDescsBkt.Cursor()
-				for pk, rawDesc := c.First(); pk != nil; pk, rawDesc = c.Next() {
+				for wantHash, rawDesc := c.First(); wantHash != nil; wantHash, rawDesc = c.Next() {
+					if len(wantHash) != 32 {
```
threebithacker, line 1435: let's use  the constant `publicKeyHashSize` here as well?

client/Makefile
```diff
@@ -3,7 +3,7 @@ ldflags="-X github.com/katzenpost/katzenpost/core/epochtime.WarpedEpoch=${warped
 uid=$(shell [ "$$SUDO_UID" != "" ] && echo "$$SUDO_UID" || id -u)
 gid=$(shell [ "$$SUDO_GID" != "" ] && echo "$$SUDO_GID" || id -g)
 docker_user=${uid}:${gid}
-docker_args=--user ${docker_user} -v $(shell readlink -f ..):/go/katzenpost -e GOCACHE=/tmp/gocache --network=host --ulimit memlock=1048576 --rm -it
+docker_args=--user ${docker_user} -v $(shell readlink -f ..):/go/katzenpost -e GOCACHE=/tmp/gocache --network=host --ulimit memlock=1048576 --rm -it --volume /usr/local/include/sphincsplus:/usr/local/include/sphincsplus
```
threebithacker, line 6: we should probably build sphincsplus inside the docker?

client/client_docker_test.go
```diff
@@ -50,63 +50,6 @@ func TestDockerClientConnectShutdown(t *testing.T) {
 	client.Wait()
 }
 
-func TestDockerClientBlockingSendReceive(t *testing.T) {
```
threebithacker, line 53: where did all these tests go? are they no longer relevant?

client/config/config.go
```diff
@@ -115,35 +117,42 @@ type NonvotingAuthority struct {
...
-		LogBackend:       l,
-		Address:          nvACfg.Address,
-		PublicKey:        nvACfg.PublicKey,
-		DialContextFn:    pCfg.ToDialContext("nonvoting:" + nvACfg.PublicKey.String()),
+		AuthorityLinkKey:     authLinkKey,
+		LinkKey:              linkKey,
+		LogBackend:           l,
+		Address:              nvACfg.Address,
+		AuthorityIdentityKey: identityPublicKey,
+		DialContextFn:        pCfg.ToDialContext(fmt.Sprintf("nonvoting: %x", linkHash[:])),
```
threebithacker, line 148: I wonder if we should consider just putting some random data here instead. Like do we care about telling Tor what our linkKey or not? If we just send some random stuff then we never have to answer that question. Thoughts?

core/wire/commands/commands.go
```diff
@@ -216,14 +221,14 @@ func getConsensusFromBytes(b []byte) (Command, error) {
 // GetVote is a de-serialized get_vote command.
 type GetVote struct {
 	Epoch     uint64
-	PublicKey *eddsa.PublicKey
+	PublicKey sign.PublicKey
 }
 
 // ToBytes serializes the GetVote and returns the resulting slice.
 func (v *GetVote) ToBytes() []byte {
 	out := make([]byte, cmdOverhead+8, cmdOverhead+voteOverhead)
```
threebithacker, line 229: not related to this PR, but `GetVote` emits a bunch of stuff that `getVoteFromBytes` doesn't parse at the beginning. I'm assuming the parsing is split across two functions somewhere. Is this intentional? Wonder we don't just use CBOR for this.
as far as I can see `cmdOverhead = 1+1+4` and `voteOverhead = 8 + PublicKeySize`, and the byte layout seems to be:
```
offset
0 : command
1 : unspecified
2: 4 bytes of voteOverhead
---- getVoteFromBytes deals only with this:
6: 8 bytes of epoch
14: whatever bytes of public key
```
do we know what `out[1]` is supposed to hold? is `command` a little-endian uint16, and the other fields are big-endian?

docker/nonvoting_mixnet/conf/mix2/mix_data/identity.public.pem
```diff
@@ -1,3 +1,4 @@
------BEGIN ED25519 PUBLIC KEY-----
-1x8tq04hY+i83o9Yn2nrfTkFj4jXIMWCWilR7fgrWyM=
------END ED25519 PUBLIC KEY-----
+-----BEGIN ED25519 SPHINCS+ PUBLIC KEY-----
```
threebithacker, line 1: hmmmmmm. I wonder if it would make sense to keep the test data for `eddsa` so it's still technically possible to run the docker tests without the pq stuff?

docker/nonvoting_mixnet/docker-compose.yml
```diff
@@ -85,6 +90,7 @@ services:
     user: $DOCKER_USER
     volumes:
       - ./conf/auth:/conf
+      - /opt/sphincsplus:/opt/sphincsplus
```
threebithacker, line 93: `/opt/sphincsplus` and `/usr/local/sphincsplus` etc -- I thought we ditched `/opt` ? Either way, why does it need to be mounted as a docker volume? Shouldn't we just build it inside the docker build ?

threebithacker (changes requested):

authority/voting/server/state_test.go
```diff
@@ -176,68 +177,77 @@ func TestVote(t *testing.T) {
...
+		topology2.Layers[1].Nodes = []config.Node{config.Node{IdentityPublicKeyPem: idKeys[2].identityPublicKeyPem},
+			config.Node{IdentityPublicKeyPem: idKeys[3].identityPublicKeyPem}}
+		topology2.Layers[2].Nodes = []config.Node{config.Node{IdentityPublicKeyPem: idKeys[5].identityPublicKeyPem},
+			config.Node{IdentityPublicKeyPem: idKeys[4].identityPublicKeyPem}}
+
+	*/
 	// one auth uses the conflicting topology, so we shall expect consensus with 2/3
 	authCfgs[0].Topology = &topology
-	authCfgs[1].Topology = &topology2
+	authCfgs[1].Topology = &topology
```
threebithacker, line 218: we changed this test to target full consensus instead of a limited threshold. we should remove the `conflicting topology` comment above if we want to keep that.

authority/voting/server/state_test.go
```diff
@@ -247,39 +257,37 @@ func TestVote(t *testing.T) {
...
-			copy(fu[:], d.desc.IdentityKey.Bytes())
-			s.descriptors[votingEpoch][fu] = d
+			s.descriptors[votingEpoch][d.desc.IdentityKey.Sum256()] = d
 		}
 	}
 
 	// exchange votes
 	for i, s := range stateAuthority {
-		t.Logf("s.s.IdentityKey: %s", s.s.IdentityKey())
+		//t.Logf("s.s.IdentityKey: %s", s.s.IdentityKey())
```
threebithacker, line 275: maybe delete it if we don't want it, alternatively restore it

authority/voting/server/state_test.go
```diff
@@ -303,88 +311,124 @@ func TestVote(t *testing.T) {
...
+
 	// initial generation of key material for each authority
-	peersMap := make(map[[eddsa.PublicKeySize]byte]*config.AuthorityPeer)
+	peersMap := make(map[[sign.PublicKeyHashSize]byte]*config.AuthorityPeer)
 	for i := 0; i < numAuthorities; i++ {
 		cfg := new(config.Config)
 		cfg.Logging = &config.Logging{Disable: false, File: "", Level: "DEBUG"}
 		cfg.Parameters = parameters
+
+		datadir, err := os.MkdirTemp("", fmt.Sprintf("auth_%d", i))
```
threebithacker, line 375: does this stuff ever get cleaned up?

server/config/config.go
```diff
@@ -740,14 +725,7 @@ func (p *Peer) validate(datadir string) error {
...
 		}
 	}
-	var pubKey eddsa.PublicKey
-	if err := pubKey.FromString(p.IdentityPublicKey); err != nil {
-		return fmt.Errorf("Voting Peer: Invalid IdentityPublicKey: %v", err)
-	}
-	_, err := wire.NewScheme().PublicKeyFromPemFile(filepath.Join(datadir, p.LinkPublicKeyPem))
-	if err != nil {
-		return fmt.Errorf("Voting Peer: Invalid Link PublicKey PEM file: %v", err)
-	}
```
threebithacker, line 750: is this intentionally removed?

client/Makefile
david415: why? Currently we'll need to build sphincsplus in order to run many of the unit tests.

client/client_docker_test.go
david415: Yeah it seems we were testing some client features that catshadow doesn't even use and some of these tests stopped working at some point due to client bugs.

client/config/config.go
david415: let's do it in another ticket/pr

core/crypto/cert/cert.go
```diff
@@ -177,22 +239,22 @@ func Sign(signer Signer, data []byte, expiration int64) ([]byte, error) {
...
 		return nil, err
 	}
-	cert.Signatures = []Signature{
-		Signature{
-			Identity: signer.Identity(),
-			Payload:  signer.Sign(mesg),
-		},
+	cert.Signatures = make(map[[32]byte]Signature)
+	cert.Signatures[verifier.Sum256()] = Signature{
+		PublicKeySum256: verifier.Sum256(),
```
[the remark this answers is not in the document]
david415: i'll make the concrete key type cache the hash when it gets created

core/crypto/cert/cert.go
```diff
@@ -108,8 +112,51 @@ type certificate struct {
 	// Version is the certificate format version.
 	Version uint32
 
-	// Expiration is seconds since Unix epoch.
-	Expiration int64
+	// Expiration is katzenpost epoch id of the expiration.
+	Expiration uint64
```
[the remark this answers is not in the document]
david415: The name Expiration is pretty clear, no?

core/crypto/cert/cert_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+func TestBadCertificate(t *testing.T) {
+
+	signingPrivKey, signingPubKey := Scheme.NewKeypair()
+
+	current, _, _ := epochtime.Now()
+
+	certified := []byte("hello, i am a message")
+
+	certificate, err := Sign(signingPrivKey, signingPubKey, certified, current+1)
```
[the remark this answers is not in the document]
david415: current + 123 is fine. the others will give the error. now we have test for this.

client/Makefile
threebithacker: Hmm, I didn't realize we didn't do the entire build inside Docker. In that case, the `--volume` will go away when `sphincsplus.o` gets linked into the binaries by `cgo` at buildtime.

client/client_docker_test.go
threebithacker: @mixmasala what's your take?
My thinking is:
- If something doesn't work on `main`, and doesn't work on the PR branch, we shouldn't have to delete the tests to get an unrelated PR merged.
- If they do work on `main`, but not on this branch (due to unrelated bugs), we should open an issue and reference it as part of the PR thread, but not necessarily fix them.
- If they do not work on `main`, but do work on the PR branch, that sounds like we fixed something, or that we are lacking test coverage. If it's the latter, open an issue.
- If the code became irrelevant (no other callers), and we want to delete the tests, we should probably delete the exposed API code they exercise, too.

client/config/config.go
threebithacker: opened https://github.com/katzenpost/katzenpost/issues/56

core/crypto/cert/cert.go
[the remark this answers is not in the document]
threebithacker: Neither the name nor the description tells me whether `Expiration: 2` means that it's valid during epoch `2`, so I don't think it's clear at all.

It's about as clear as `11:59:59pm` referring to the time one or two seconds before `12am`, and thus `3600` or `3601` seconds before `1am` depending on what day and what version of the IERS Bulletin C your timestamp-to-string function refers to.

core/crypto/cert/cert_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+	scheme := Scheme
+	_, ephemeralPubKey := scheme.NewKeypair()
+
+	signingPrivKey, signingPubKey := scheme.NewKeypair()
+
+	current, _, _ := epochtime.Now()
+
+	toSign := ephemeralPubKey.Bytes()
+	certificate, err := Sign(signingPrivKey, signingPubKey, toSign, current+123)
```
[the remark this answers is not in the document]
david415: n=1 is the smallest n

threebithacker (changes requested):
I'm almost through reviewing the `Golang` code. I haven't even looked at `sphincsplus/`, I'm just going to pretend that does what it says on the tin.

docker/nonvoting_mixnet/conf/mix2/mix_data/identity.public.pem
david415: i don't think we want to accumulate more pem files

authority/voting/server/state_test.go
david415: the tests creates the temp dir but never cleans up. that's cool, right?

server/config/config.go
david415: yes we load it elsewhere

authority/voting/server/state.go
```diff
@@ -647,8 +668,8 @@ func (s *state) sendVoteToPeer(peer *config.AuthorityPeer, vote []byte, epoch ui
 // for our link layer wire protocol as specified by
 // the PeerAuthenticator interface in core/wire/session.go
 func (s *state) IsPeerValid(creds *wire.PeerCredentials) bool {
-	var ad [eddsa.PublicKeySize]byte
-	copy(ad[:], creds.AdditionalData)
+	var ad [publicKeyHashSize]byte
+	copy(ad[:], creds.AdditionalData[:publicKeyHashSize])
```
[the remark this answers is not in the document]
david415: you want it to err if there is more data?

client/session.go
```diff
@@ -125,8 +125,8 @@ func NewSession(
 	if err != nil {
 		panic(err)
 	}
-	if len(linkKeyText) > 64 {
-		linkKeyText = linkKeyText[:64]
+	if len(linkKeyText) > 32 {
+		linkKeyText = linkKeyText[:32]
```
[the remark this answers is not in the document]
david415: yes but currently "username" is a string... i don't want to change client or minclient too much here in this PR

core/crypto/sign/ed25519sphincsplus/scheme.go
```diff
@@ -0,0 +1,207 @@
...
+func (p *publicKey) FromBytes(data []byte) error {
+	if len(data) != eddsa.PublicKeySize+sphincs.PublicKeySize {
+		return ErrPublicKeySize
+	}
+	err := p.e.FromBytes(data[:eddsa.PublicKeySize])
+	if err != nil {
+		return err
+	}
+	err = p.s.FromBytes(data[eddsa.PublicKeySize:])
+	if err != nil {
```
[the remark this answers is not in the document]
david415: it's fine. always check for err.

server/internal/outgoing/connector.go
```diff
@@ -173,12 +173,25 @@ func (co *connector) onClosedConn(c *outgoingConn) {
 }
 
 func (co *connector) IsValidForwardDest(id *[constants.NodeIDLength]byte) bool {
+
+	co.log.Debug("IsValidForwardDest start")
```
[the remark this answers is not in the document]
david415: removed

authority/voting/server/state.go
```diff
@@ -1459,46 +1494,102 @@ func newState(s *Server) (*state, error) {
...
 
 	// Initialize the authorized peer tables.
-	st.authorizedMixes = make(map[[eddsa.PublicKeySize]byte]bool)
+	st.reverseHash = make(map[[publicKeyHashSize]byte]sign.PublicKey)
+	st.authorizedMixes = make(map[[publicKeyHashSize]byte]bool)
 	for _, v := range st.s.cfg.Mixes {
-		pk := v.IdentityKey.ByteArray()
+		_, identityPublicKey := cert.Scheme.NewKeypair()
+		if filepath.IsAbs(v.IdentityPublicKeyPem) {
+			err := pem.FromFile(v.IdentityPublicKeyPem, identityPublicKey)
```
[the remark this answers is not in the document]
david415: could do or not either way

docker/nonvoting_mixnet/docker-compose.yml
david415: fixed

docker/nonvoting_mixnet/conf/mix2/mix_data/identity.public.pem
david415: also it's not possible to run docker with the old keying unless you change cert.go to use the old scheme

core/wire/commands/commands.go
david415: this hasn't changed that much in this PR.

core/crypto/pem/pem_test.go
```diff
@@ -24,18 +24,15 @@ import (
...
-	privKey, err := NewScheme().Load(privFile, pubFile, rand.Reader)
-	require.NoError(t, err)
+func TestToFromPEM(t *testing.T) {
+
+	datadir := os.TempDir()
 
-	privKey2, err := NewScheme().Load(privFile, pubFile, nil)
+	idKey, _ := cert.Scheme.NewKeypair()
+	err := ToFile(filepath.Join(datadir, "identity.private.pem"), idKey)
 	require.NoError(t, err)
```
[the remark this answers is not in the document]
david415: added tests

core/crypto/sign/eddsa/eddsa_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	pubKey2, err := Scheme.UnmarshalTextPublicKey(pubKeyText)
+	require.NoError(t, err)
+
+	signature := privKey.Sign(message)
+	ok := pubKey.Verify(signature, message)
+	require.True(t, ok)
+
+	ok = pubKey2.Verify(signature, message)
+	require.True(t, ok)
```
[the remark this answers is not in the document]
david415: we test it in each module

core/crypto/sign/eddsa/eddsa_test.go
```diff
@@ -0,0 +1,52 @@
...
+	privKey, pubKey := Scheme.NewKeypair()
+
+	pubKeyText, err := pubKey.MarshalText()
+	require.NoError(t, err)
+
+	pubKey2, err := Scheme.UnmarshalTextPublicKey(pubKeyText)
+	require.NoError(t, err)
+
+	signature := privKey.Sign(message)
+	ok := pubKey.Verify(signature, message)
```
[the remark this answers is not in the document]
david415: well this is just a wrapper around core/crypto/eddsa .... which is adequately tested.

client/client_docker_test.go
david415: I mean... I've said I have plans to fix the client library. Can't do everything in this PR.

README.rst
```diff
@@ -11,6 +11,30 @@ Visit the project website for more information about Katzenpost.
...
+As part of this change in our cryptographic protocols we have added the use
+of Sphincs+, the stateless hash based post quantum signature scheme.
+
+The katzenpost monorepo now has a copy of the sphincsplus git repo in `katzenpost/sphincsplus`.
+
+Install this dependency like this:
+
+```
+cd katzenpost/sphincsplus/ref
+make libsphincsplus.so
```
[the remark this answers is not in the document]
david415: https://github.com/katzenpost/katzenpost/issues/57

authority/nonvoting/server/state.go
```diff
@@ -615,9 +620,9 @@ func (s *state) restorePersistence() error {
 				}
 
 				c := eDescsBkt.Cursor()
-				for pk, rawDesc := c.First(); pk != nil; pk, rawDesc = c.Next() {
-					verifier := new(eddsa.PublicKey)
-					err := verifier.FromBytes(pk)
+				for wantHash, rawDesc := c.First(); wantHash != nil; wantHash, rawDesc = c.Next() {
+					_, verifier := cert.Scheme.NewKeypair()
+					err := verifier.FromBytes(wantHash)
```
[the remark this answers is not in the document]
david415: yeah that looks like a bug... and it only affects nonvoting so that's what i didn't notice it yet

core/crypto/sign/ed25519sphincsplus/scheme.go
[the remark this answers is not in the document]
threebithacker: scenario would be where you have a key, try to overwrite it with another key, get an error, and now the old key holds a mix of the two keys.

authority/nonvoting/server/state.go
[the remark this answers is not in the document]
david415: fixed

threebithacker (changes requested):
OK, I have read through all the changes and I feel pretty good about this now.

client/session.go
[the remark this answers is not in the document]
threebithacker: `fmt.Sprintf("%x", s.linkKey.PublicKey().Sum256())` ?

authority/voting/server/state.go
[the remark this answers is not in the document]
threebithacker: Yeah, I think so. There's enough things that depend on that property being true.

authority/voting/server/state_test.go
threebithacker: (to be sure that is a lot better than the hardcoded paths that were there before)

docker/nonvoting_mixnet/conf/mix2/mix_data/identity.public.pem
threebithacker: OK, fair enough.

core/wire/commands/commands.go
threebithacker: no, it's unrelated to that this PR really.

mixmasala:
Why are you merging broken code into master without review again?

client/session.go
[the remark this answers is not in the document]
mixmasala: Is this resolved?

PR #49, opened 2022-10-12
Replace cbor encoder

mixmasala:
Where possible I replaced the cbor encoder that we were using, as per:
https://github.com/katzenpost/core/issues/103


david415:
lgtm

david415 (approved):

PR #68, opened 2022-11-10
minclient: fix deadlock onConnStatusChange and getConsensus

mixmasala:
This PR is created to track proposed fixes to minclient that remove deadlock on the connection Lock

### DIFF
--- a/minclient/connection.go
+++ b/minclient/connection.go
@@ -695,6 +695,7 @@ func (c *connection) sendPacket(pkt []byte) error {
 		c.Unlock()
 		return ErrNotConnected
 	}
+	c.Unlock()
 
 	errCh := make(chan error)
 	c.sendCh <- &connSendCtx{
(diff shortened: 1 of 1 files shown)


david415 (approved):

minclient/connection.go
```diff
@@ -717,6 +715,7 @@ func (c *connection) getConsensus(ctx context.Context, epoch uint64) (*commands.
 		c.Unlock()
 		return nil, ErrNotConnected
 	}
+	c.Unlock()
```
threebithacker, line 718: since we're repeating this, wonder if it would make sense to have a method like
```go
func (c *connection) isConnectedMethod() (err) {
  c.Lock()
  if !c.isConnected {
    err = ErrNotConnected
  }
  c.Unlock()
}
```

threebithacker (approved):

minclient/connection.go
mixmasala: I dont feel that results in signifcant deduplication, while adding a small overhead

PR #65, opened 2022-11-05
add end to end integration tests to github workflow

mixmasala:
this PR adds end to end tests

PR #69, opened 2022-11-11
Fix make clean

mixmasala:
Adding the -x flag to git clean removes the untracked files from the testnet directories, reclaiming space used by logfiles and ensuring a clean start between builds.

```
       -x
           Don\u2019t use the standard ignore rules (see gitignore(5)), but still
           use the ignore rules given with -e options from the command line.
           This allows removing all untracked files, including build products.
           This can be used (possibly in conjunction with git restore or git
           reset) to create a pristine working directory to test a clean
           build.
```


mixmasala:
test_e2e will not complete because that workflow doesn't exist until #65 is merged

threebithacker (changes requested):
looks like this is based on top of [4ad1a8b](https://github.com/katzenpost/katzenpost/pull/69/commits/4ad1a8b67fa8219a0590b8481559cee2619e0eef) instead of `main` ?

mixmasala:
> looks like this is based on top of [4ad1a8b](https://github.com/katzenpost/katzenpost/pull/69/commits/4ad1a8b67fa8219a0590b8481559cee2619e0eef) instead of `main` ?

#65 #68

PR #74, opened 2022-11-21
Remove inline prometheus

mixmasala:
This is based on top of #65 and disables the prometheus tests unless enabled with build tag promethius


david415:
why are there so many changes to fix this one little bug?

PR #79, opened 2022-11-23
Add makefile targets

mixmasala:
This adds makefile targets to server and normalize key names


david415:
do you want to keep this and make it work?

mixmasala:
no, things have changed too much, so it's not worth my time to fix this

PR #81, opened 2022-11-23
Add pem helper

mixmasala:
This PR adds a helper to authority so that PEM files or PEM-encoded strings can be used with the configuration.

I might need to add another helper for the mix server as well.


david415:
can we instead open a ticket for this task?

ioerror:
Ideally it would be possible to define the pki as a series of files rather than PEM strings in the dirauth and mix config files. This is already the case for mix and provider id public keys.

Issue #83, opened 2022-11-24
restarting authority fails because old certificate from persistence db is expired


mixmasala:
I would prefer to not use on-disk persistence because I suspect it might result in conflicting descriptors upon restart that might cause an authority to have a wrong view of the network when making consensus.

mixmasala:
the old authority code used the following value for certificate expiration:
-       CertificateExpiration = (epochtime.Period * 3) + (time.Minute * 10)
so that a certificate is valid for all the epochs its keys are valid + some timeslop
I think partly we're too aggressive on the certificate expiration

mixmasala:
+       signed, err := cert.Sign(signer, verifier, payload, epoch+2)

is the new value

mixmasala:
probably that should actually be epoch+3 because the keys in a consensus are stil valid and the descriptor is being signed in the epoch before it is valid

david415:
fix here https://github.com/katzenpost/katzenpost/pull/104

PR #84, opened 2022-11-24
Extend MixDescriptor certificate expiration to 3 epochs

mixmasala:
The next epoch the descriptor becomes valid, and contains 3 epochs worth of MixKeys, so it should be valid for current epoch + 4

See #83


mixmasala:
Also do the same thing for the consensus

david415:
lgtm

david415 (approved):

PR #88, opened 2022-11-25
Fix whitelist bug

mixmasala:
This adds some extra verbose logging to the voting authority and fixes a bug where MixDescriptors sent by another authority were included in consensus without checking that they were whitelisted.


ioerror:
Lets call it allowlist, rather than whitelist.

mixmasala:
> Lets call it allowlist, rather than whitelist.

patches welcome :)

PR #97, opened 2022-11-28
Add threat model doc work in progress

david415:


mixmasala (approved):
Could you elaborate on what the distinction is between a timing attack and statistical disclosure attack? 

for section 6, perhaps DoS in conjunction with n-1 or statistical disclosure attacks speeds up the attack, e.g. a hybrid attack could be explained, and the adversary in the case of DoS is one who sees some part of the network and has capacity to impact some other part of the network.

What is a timing attack essentially? Do we consider active traffic manipulation / confirmation to be a kind of timing attack? Or do we mean something like sidechannels?
For example, an adversary who can interfere with client connections to provider nodes and can also observe some part of the network, e.g. a provider service, in order to confirm a specific client is using the service. For example, disconnecting a node and then observing a decrease in requests / decrease in latency of request being served used as a confirmation.

I think we can merge this now, but sections 5,6 should be expanded at least to describe the essential elements of the attack.

Issue #99, opened 2022-11-29
don't truncate the sphinx MAC


david415:

ioerror:
This is now resolved by the merge of #101.

PR #101, opened 2022-11-29
Pq sphinx

david415:
THIS PR SOLVES BOTH https://github.com/katzenpost/katzenpost/issues/102 AND https://github.com/katzenpost/katzenpost/issues/99


ioerror (approved):
This likely requires a flag day.

Issue #103, opened 2022-11-29
voting authority BUG: persisted certificate is expired


david415:
Currently, operators of the voting authority are in the habit of rm`ing the persistence.db file which the voting authority creates. This is because the certificate saved to that file on disk will have an expired epoch due to a off by one or off by a few error. It would be advisable to fix this as soon as possible.

david415:
duplicate ticket https://github.com/katzenpost/katzenpost/issues/83

PR #104, opened 2022-11-29
Fix voting auth cert db bug

david415:
Fix bug describe in this ticket here https://github.com/katzenpost/katzenpost/issues/103


ioerror (approved):
This requires a flag day.

Issue #11, opened 2021-09-10
Add provider config to distinguish between entry node and service provider


david415:
We should perhaps disallow entry nodes to coexist with service provider on the same host. Instead the approach we want is to have some providers hosting mixnet services like memspool and other providers acting as entry nodes for the mixnet client connections. I reckon it would be a simple code change.

The open question for katzenpost dev team is: Should we make the above mandatory or optional?

mixmasala:
```
06:21:12.744 DEBU minclient/conn:W3MJY7GLp+4yIp6K0yFWFSC4gFIS8fN+lcc9zRWCNRU=@provider1: PKI fetch successful.
06:21:12.745 DEBU minclient/conn:W3MJY7GLp+4yIp6K0yFWFSC4gFIS8fN+lcc9zRWCNRU=@provider1: Descriptor for epoch 116527: &{Name:provider1 IdentityKey:2krwfNDfbakZCSTUUZYKXwdduzlEgS9Jfwm7eyZ0sCg= LinkKey:bAAHyaf/fl682PBj2VWzGRMVfWBV5t+oBaaKUg+6ZhE= MixKeys:map[116527:zHRcV2dookTuQcImGMt8MSBVBXMUwLR/uM3IXmjoikU= 116528:VWqAJ3pm0KmQnyCBvq0qJvp9NVdubU+m5szwpe6vkAk= 116529:BgJ01aJqj0OoqVBukmNPPJmfaF5LpHgTOgaeV9h0iys=] Addresses:map[tcp4:[172.28.1.3:59484]] Kaetzchen:map[echo:map[endpoint:echo] panda:map[endpoint:+panda] spool:map[endpoint:+spool]] Layer:255 LoadWeight:0 AuthenticationType:tofu}
06:21:12.746 DEBU minclient/conn:W3MJY7GLp+4yIp6K0yFWFSC4gFIS8fN+lcc9zRWCNRU=@provider1: Dialing: 172.28.1.3:59484
06:21:12.747 DEBU minclient/conn:W3MJY7GLp+4yIp6K0yFWFSC4gFIS8fN+lcc9zRWCNRU=@provider1: TCP connection established.
[...]

PR #107, opened 2022-11-30
Entry vs service mix

david415:
satisfy this ticket: https://github.com/katzenpost/katzenpost/issues/11

### DIFF
--- a/docker/nonvoting_mixnet/conf/provider1/katzenpost.toml
+++ b/docker/nonvoting_mixnet/conf/provider1/katzenpost.toml
@@ -59,28 +59,26 @@
 #
 
 [Provider]
-EnableEphemeralClients = true
-TrustOnFirstUse = true
-
-   [[Provider.CBORPluginKaetzchen]]
+  EnableServiceProvider = true
+  [[Provider.CBORPluginKaetzchen]]
     Capability = "echo"
-    Endpoint = "echo"
+    Endpoint = "+echo"
     Command = "/go/bin/echo_server"
     MaxConcurrency = 1
     [Provider.CBORPluginKaetzchen.Config]
-      log_level = "DEBUG"
-      log_dir = "/conf/service_logs"
+     log_level = "DEBUG"
+     log_dir = "/conf/service_logs"
 
-   [[Provider.CBORPluginKaetzchen]]
-     Disable = false
-     Capability = "panda"
-     Endpoint = "+panda"
-     Command = "/go/bin/panda_server"
-     MaxConcurrency = 1
-     [Provider.CBORPluginKaetzchen.Config]
-      log_dir = "/conf/service_logs"
-      log_level = "DEBUG"
-      fileStore = "/conf/service_data/panda.storage"
+  [[Provider.CBORPluginKaetzchen]]
+    Disable = false
+    Capability = "panda"
+    Endpoint = "+panda"
+    Command = "/go/bin/panda_server"
+    MaxConcurrency = 1
+    [Provider.CBORPluginKaetzchen.Config]
+     log_dir = "/conf/service_logs"
+     log_level = "DEBUG"
+     fileStore = "/conf/service_data/panda.storage"
 
   [[Provider.CBORPluginKaetzchen]]
     Disable = false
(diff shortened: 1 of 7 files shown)


docker/nonvoting_mixnet/conf/provider1/katzenpost.toml
```diff
@@ -59,28 +59,26 @@
...
 [Provider]
-EnableEphemeralClients = true
-TrustOnFirstUse = true
-
-   [[Provider.CBORPluginKaetzchen]]
+  EnableServiceProvider = true
+  [[Provider.CBORPluginKaetzchen]]
     Capability = "echo"
-    Endpoint = "echo"
+    Endpoint = "+echo"
```
mixmasala, line 65: what was happening prior to adding +?

mixmasala (changes requested):
The code changes ooks good to me, except I am not sure that we want to remove the loop service from the entry provider - how does this affect cover traffic distribution?

I'm not sure if we need to change this behavior, but let's discuss implications.

docker/nonvoting_mixnet/conf/provider1/katzenpost.toml
david415: it doesn't matter what we call it. it's just a string in a database

david415: no. what I meant was that this change on this line isn't needed for things to work. It works if it's set to "echo" and also works if set to "+echo". I hope that this is clear now. The endpoint was being found and there was no reason to think it wasn't. Nothing wrong with this line if we do not change it. Jesus.

ioerror:
It can be slightly confusing to run a service mix that does not take client connections.  I had confusion with this configuration of the mix service:
```
#  EnableEphemeralClients = true # this is false by default, if left unset
  TrustOnFirstUse = true
```

I first understood that `EnableEphemeralClients` is false by default (and unset, or it could be explicitly set to false by a config) and as a result, I understood that the `TrustOnFirstUse` value would only apply to connections that weren't from clients. It seems that this isn't true. Both need to be false or client connections will be accepted.

Having a single overarching flag such as `AllowClients = true`  would mean that `EnableServiceProvider = true` could allow the config to be treated as an error at startup time. Similarly, `AllowClients = false` could cause an error if `EnableEphemeralClients`  or `TrustOnFirstUse` were set (and would otherwise allow clients to connect, which is not what anyone here wants).

mixmasala:
> It can be slightly confusing to run a service mix that does not take client connections. I had confusion with this configuration of the mix service:
> 
> ```
> #  EnableEphemeralClients = true # this is false by default, if left unset
>   TrustOnFirstUse = true
> ```
> 
> I first understood that `EnableEphemeralClients` is false by default (and unset, or it could be explicitly set to false by a config) and as a result, I understood that the `TrustOnFirstUse` value would only apply to connections that weren't from clients. It seems that this isn't true. Both need to be false or client connections will be accepted.
> 

This is backwards. TrustOnFirstUse should only apply to client connections.
The documentation merely states:
https://github.com/katzenpost/katzenpost/blob/main/docs/handbook/mix_server.rst

```
EnableEphemeralClients if set to true allows ephemeral clients to be created when the Provider first receives a given user identity string.
TrustOnFirstUse if set to true the Provider will trust client's wire protocol keys on first use.
```

And the docstring comments for those fields say:
https://github.com/katzenpost/katzenpost/blob/c185bfd95f21ffed1521c8fc6b52e0ea6df01f1c/server/config/config.go#L363
```
// TrustOnFirstUse indicates whether or not to trust client's wire protocol keys
// on first use. If set to true then first seen keys cause an entry in the userDB
// to be created. It will later be garbage collected.

// EnableEphemeralhClients is set to true in order to
// allow ephemeral clients to be created when the Provider
// first receives a given user identity string.
```

It should be clear that the TrustOnFirstUse option applies to client connections.
What is not so clear is that EnableEphemeralClients is also required for the TOFU authentication method to be announced in the PKI, and that EnableEphemeralAccounts turns on garbage collection, which is not true about the first option if EnableEphemeralAccounts is not enabled.

> Having a single overarching flag such as `AllowClients = true` would mean that `EnableServiceProvider = true` could allow the config to be treated as an error at startup time. Similarly, `AllowClients = false` could cause an error if `EnableEphemeralClients` or `TrustOnFirstUse` were set (and would otherwise allow clients to connect, which is not what anyone here wants).

Some operators may prefer to use a different authentication mode for their users using the userdb rather than allow or disallow all client connections. EnableServiceProvider means more than just accepting client connections, as it also enables provider services. Similarly, they may want new accounts to be able to create a userdb entry and continue to keep that identity, rather than 

The docstring for EnableEphemeralClients is not terribly clear what it actually does, which is enable garbage collection for message queues and used as a conditional to enable publishing the TOFU option. I think that's wrong, it shouldn't be needed to enable publishing the PKI field, and the description of TrustOnFirstUse should not say that accounts will be garbage collected because that option doesn't in fact enable the garbage collection timer.

My preference is to consolidate the options EnableEphemeralAccounts and TrustOnFirstUse so that

mixmasala (approved):

evainfeld:
I'm gonna answer the cover traffic-loops thing in MM

david415:
this is a duplicate. working on it here https://github.com/katzenpost/katzenpost/pull/520

Issue #108, opened 2022-11-30
update wire protocol doc


david415:
pqXX with the hybrid KEM Kyber768 + X25519

PR #109, opened 2022-11-30
Update noise wire protocol spec

david415:
update the wire protocol spec so it accurately represents what we are now using. 
fixes this ticket: https://github.com/katzenpost/katzenpost/issues/108


ioerror (approved):

ioerror:
It would be good if the two (canceled, failing) tests pass. The spec change seems reasonable.

PR #114, opened 2022-12-05
Refactor deduplicate authority config types

mixmasala:
This changes the authority configuration so as to remove the [Authority] key, and in so doing, use authority/voting/server/config.Authority consistently , removing types config.AuthorityPeer, server/config.Peer, and the function AuthorityPeersFromPeers.


mixmasala:
Possibly [Authority] should be replaced with a section [Server] instead, so as to be somewhat more consistent with the mix server configuration as well?

PR #126, opened 2022-12-12
authority: print offending node id in PEM format

david415:


ioerror (approved):

leif:
in the soon-to-be-merged #148 we're logging node names and key hashes here now instead.

PR #127, opened 2022-12-12
wip_fix_commit_and_reveal

mixmasala:
this is work in progress to fix an issue where an authority can commit to different values in each vote and break the consensus


mixmasala:
so, this still needs signature verification, and we should consider if there is a smarter way to solve this problem.

david415 (commented):
looks good so far. although i think we should use hmac.Equal instead of bytes.Equal but that's a minor thing.

david415 (approved):

Issue #134, opened 2022-12-21
replace AVL tree with priority queue in decoy traffic module of mix server


david415:
AVL tree library has a bug... let's just not use an AVL tree. Priority  queue is fine for this task most likely. Otherwise if it must be an AVL tree (i highly doubt it) then we fix the bug in the AVL tree.

threebithacker:
I believe we fixed this in https://github.com/katzenpost/katzenpost/pull/221 - reopen if you disagree @david415

Issue #133, opened 2022-12-20
omelas mix crashes 


ioerror:
Mix server omelas appears to have had two crashes. This happened around the same time that nutria's disk filled. Nutria crashed which caused nutria to be unreachable. I think that the mix server is using tag v0.0.15.

```
panic: interface conversion: interface {} is *decoy.surbCtx, not []*decoy.surbCtx

goroutine 36 [running]:
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).sweepSURBCtxs(0xc000000000)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:432 +0x57e
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).worker(0xc000000000)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:190 +0x53a
github.com/katzenpost/katzenpost/core/worker.(*Worker).Go.func1()
        /home/katzenpost/katzenpost/core/worker/worker.go:39 +0x5a
created by github.com/katzenpost/katzenpost/core/worker.(*Worker).Go
        /home/katzenpost/katzenpost/core/worker/worker.go:37 +0xb9
```

```
panic: BUG: SURB mismatch in node ETA list

goroutine 37 [running]:
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).loadAndDeleteSURBCtx(0xc0000ac500, 0x5?)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:408 +0x251
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).OnPacket(0xc0000ac500, 0xc010c81780)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:110 +0x279
github.com/katzenpost/katzenpost/server/internal/cryptoworker.(*Worker).worker(0xc0000c0380)
        /home/katzenpost/katzenpost/server/internal/cryptoworker/crypto_worker.go:271 +0xead
github.com/katzenpost/katzenpost/core/worker.(*Worker).Go.func1()
        /home/katzenpost/katzenpost/core/worker/worker.go:39 +0x5a
created by github.com/katzenpost/katzenpost/core/worker.(*Worker).Go
        /home/katzenpost/katzenpost/core/worker/worker.go:37 +0xb9
```

PR #135, opened 2022-12-21
fix for half of #133

david415:
This PR fixes first crash bug from issue https://github.com/katzenpost/katzenpost/issues/133
It is a workaround style bug fix where we know for sure (because of gaurantees of golang type system) that the bug is in the AVL library. See ticket to replace with priority queue here: https://github.com/katzenpost/katzenpost/issues/134


ioerror (approved):
LGTM

PR #143, opened 2022-12-24
core/wire: fix KEM key unmarshaling

david415:
This fixes a stupid cryptography bug where i was calling generate key in each key unmarshal. But it turns out we don't have to do that because the upstream library, nyquist provides the parse methods. so it was my fault all along. sorry about that.

### DIFF
--- a/core/wire/kem.go
+++ b/core/wire/kem.go
@@ -19,13 +19,13 @@ package wire
 
 import (
 	"bytes"
-	"crypto/rand"
 	"encoding"
 	"encoding/base64"
 	"encoding/pem"
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/katzenpost/nyquist/kem"
 	"github.com/katzenpost/nyquist/seec"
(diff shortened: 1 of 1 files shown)


threebithacker (approved):

core/wire/kem.go
```diff
@@ -295,59 +295,86 @@ type scheme struct {
...
+	buf, err := os.ReadFile(f)
 	if err != nil {
 		return nil, err
 	}
-	return privKey, nil
+	blk, _ := pem.Decode(buf)
+	if blk == nil {
+		return nil, fmt.Errorf("failed to decode PEM file %v", f)
+	}
+	if blk.Type != keyType {
```
threebithacker, line 330: this should be `strings.ToUpper(blk.Type)` because the keytype is `Kyber768-X25519` in a bunch of places and `KYBER768-X25519` in other places, so this fails to read e.g. the keys in `client/testdata/client.toml`

threebithacker (changes requested):

core/wire/kem.go
threebithacker: same issue for the public keys I'm guessing?

core/wire/kem.go
```diff
@@ -295,59 +295,86 @@ type scheme struct {
...
+	buf, err := os.ReadFile(f)
 	if err != nil {
 		return nil, err
 	}
-	return pubKey, nil
+	blk, _ := pem.Decode(buf)
+	if blk == nil {
+		return nil, fmt.Errorf("failed to decode PEM file %v", f)
+	}
+	if blk.Type != keyType {
```
threebithacker, line 350: same issue here I think

threebithacker (changes requested):

core/wire/kem.go
david415: done

core/wire/kem.go
david415: done

threebithacker (approved):

PR #145, opened 2022-12-24
Add wire kem unit tests

david415:


core/wire/kem.go
```diff
@@ -129,39 +125,11 @@ func (p *publicKey) KeyType() string {
...
-		return fmt.Errorf("attempted to serialize scrubbed key")
-	}
-	blk := &pem.Block{
-		Type:  keyType,
-		Bytes: p.Bytes(),
-	}
-	return os.WriteFile(f, pem.EncodeToMemory(blk), 0600)
-}
-
 // XXX FIXME
```
threebithacker, line 128: we can change `XXX FIXME` or get rid of it, right?

core/wire/kem_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	_, err = DefaultScheme.PublicKeyFromPemFile("definitelydoesnotexist")
+	require.Error(t, err)
+
+	wrongPemData := `-----BEGIN ED25519 SPHINCS+ PUBLIC KEY-----
+2JQzwEwGBxBkQ0quWab4MD2T3E/WozBsfAMzp9wDLDaSm4jMADucY0gHAxX08iM3
+e/o5l00d9jhM5Gr51yY5FT8acP8IdPeDS1ccwW1HTpmAWMQOJyZwvo9jwiog9IVq
+-----END ED25519 SPHINCS+ PUBLIC KEY-----
+`
+	wrongPemPath := filepath.Join(os.TempDir(), "wrongpubkey1.pem")
```
threebithacker, line 53: instead of `os.TempDir()` I think we want to use a single call to this [testing.TempDir()](https://pkg.go.dev/testing#T.TempDir) per unique directory we need. That way it will get deleted after the test runs, so we don't litter `/tmp` with tons of stuff.

threebithacker (changes requested):

PR #146, opened 2022-12-24
move obsolete spec file into old dir

david415:
move aside an obsolete spec file


threebithacker (approved):

threebithacker:
What's the status of mailproxy? `docs/handbook/mailproxy.rst` talks about mailproxy, is that also obsolete?

PR #166, opened 2022-12-28
merge fixes to authority from devel

mixmasala:
this merges fixes to the voting authority protocol where authorities were re-assigning nodes to the topology but populating the list of node from a map and not sorting the list afterwards, and an issue where the priorsharedrandom was not set correctly, and also updates the project dependencies to their latest versions.


mixmasala:
I found another issue, which is that the generateTopology method is shuffling mixes when reampping the topology from the last consensus. This isn't correct behavior - the topology should remain stable if the mixes haven't changed.

david415 (approved):

leif (approved):

PR #167, opened 2022-12-29
misc cleanup

leif:
this removes some outdated comments, extra slashes at the beginning of some paths, and unnecessary apt install commands in the ci.

it also moves some variables around in the docker file.


mixmasala (approved):
looks like this is working

PR #168, opened 2022-12-29
genconfig: generate a static topology

leif:
This makes genconfig able to (optionally) include a static topology in authority configs.

It also adds options for specifying alternate parameters via a TOML file, and for specifying the number of layers.

This is based on top of #167 so the diff will be smaller if that is merged first.


leif:
oops, sorry, this isn't ready yet.

leif:
i am closing this to open a new PR because github is still showing commits here which were already merged in the previous PR

PR #187, opened 2023-01-04
server: pkiclient refetch according to voting scheule

mixmasala:
adjust the pki wakeup timer so that it sleeps until the pki publication deadline

### DIFF
--- a/server/internal/pki/pki.go
+++ b/server/internal/pki/pki.go
@@ -220,7 +220,23 @@ func (p *pki) worker() {
 			}
 		}
 
-		timer.Reset(recheckInterval)
+		now, elapsed, till := epochtime.Now()
+
+		// it's after the consensus publication deadline
+		if elapsed > vServer.PublishConsensusDeadline {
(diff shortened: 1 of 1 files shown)


mixmasala:
seems to reliably fail the client with ErrNoDocument

mixmasala:
> seems to reliably fail the client with ErrNoDocument

I forgot to wake up the pkiclient in time to publish the mix descriptor, if applicable. fixed in ddfa676

server/internal/pki/pki.go
```diff
@@ -220,7 +220,32 @@ func (p *pki) worker() {
 			}
 		}
```
david415, line 222: it's not ideal to make this worker method even longer

david415 (approved):
lgtm

server/internal/pki/pki.go
mixmasala: refactored into a method

mixmasala:
I made one change which was to move the timer update into a method per feedback in review. Thanks!

PR #189, opened 2023-01-06
Client2: the new improved mixnet client

david415:
ready for review. do not merge. work-in-progress.

### DIFF
--- a/docs/specs/wire-protocol.md
+++ b/docs/specs/wire-protocol.md
@@ -35,14 +35,11 @@ as opaque byte vectors.
 
 `x | y` denotes the concatenation of x and y.
 
-### 1.2 Kyber Key Encapsulation Mechanism
+### 1.2 Key Encapsulation Mechanism
 
-This protocol uses the Kyber Key Encapsulation Mechanism
-[KYBER](#KYBER) (NIST round 3). Kyber is one of the
-finalists in the NIST post-quantum cryptography project. Please see
-the Kyber project page for more information:
-
-- https://pq-crystals.org/kyber/
+This protocol uses ANY Key Encapsulation Mechanism.  However it's
+recommended that most users select a hybrid post quantum KEM such as
+Xwing. [XWING](#XWING)
 
 ## 2. Core Protocol
 
@@ -56,10 +53,11 @@ Our transport protocol begins with a prologue, Noise handshake,
 followed by a stream of Noise Transport messages in a minimal framing
 layer, over a TCP/IP connection.
 
-Our Noise protocol string:
+Our Noise protocol is configurable via the KEM selection in the TOML configuration files,
+here's an example PQ Noise protocol string:
 
 ```
-Noise_pqXX_Kyber768X25519_ChaChaPoly_BLAKE2b
+Noise_pqXX_Xwing_ChaChaPoly_BLAKE2b
 ```
 
 The protocol string is a very condensed description of our protocol.
@@ -70,8 +68,8 @@ pqXX: -> e <- ekem, s -> skem, s <- skem
 ```
 
 The next part of the protocol string specifies the KEM,
-`Kyber768X25519` which is a hybrid KEM where the share
-secret outputs of both X25519 and Kyber768 are combined.
+`Xwing` which is a hybrid KEM where the share
+secret outputs of both X25519 and MLKEM768 are combined.
 
 Finally the `ChaChaPoly_BLAKE2b` parts of the protocol
 string indicate which stream cipher and hash function we are using.
@@ -82,7 +80,7 @@ messages over our Noise protocol because of our using the Sphincs+
 signature scheme which has signatures that are about 49k bytes.
 
 It is assumed that all parties using the KMNWP protocol have a fixed long or
-short lived `Kyber768X25519` keypair ([KYBER](#KYBER) and [RFC7748](#RFC7748)),
+short lived `Xwing` keypair [XWING](#XWING),
 the public component of which is known to the other party in advance. How such
 keys are distributed is beyond the scope of this document.
 
@@ -274,7 +272,49 @@ is the Sphinx Packet destined for the responder.
 Initiators MUST terminate the session immediately upon reception of a
 `send_packet` command.
 
-## 4. Anonymity Considerations
+## 4. Command Padding
+
+We use traffic padding to hide from a passive network observer which
+command has been sent or received. 
+
+Among the set of padded commands we exclude the `Consensus` command because
+it's contents are a very large payload which is usually many times larger
+than our Sphinx packets. Therefore we only pad these commands:
+
+GetConsensus
+NoOp
+Disconnect
+SendPacket
+RetrieveMessage
+MessageACK
+Message
+MessageEmpty
+
+However we split them up into two directions,
+client to server and server to client because they differ in size
+due to the difference in size between `SendPacket` and `Message`:
+
+Client to Server commands:
+
+NoOp
+SendPacket
+Disconnect
+RetrieveMessage
+GetConsensus
+
+Server to client commands:
+
+Message
+MessageACK
+MessageEmpty
+
+The `GetConsensus` command is a special case because we only want to pad
+it when it's sent over the mixnet. We don't want to pad it when
+sending to the dirauths. Although it would not be so terrible if it's
+padded when sent to the dirauths... it would just needlessly take up
+bandwidth without providing any privacy benefits.
+
+## 5. Anonymity Considerations
 
 Adversaries being able to determine that two parties are communicating
 via KMNWP is beyond the threat model of this protocol. At a minimum,
@@ -282,7 +322,7 @@ it is trivial to determine that a KMNWP handshake is being performed,
 due to the length of each handshake message, and the fixed positions
 of the various public keys.
 
-## 5. Security Considerations
+## 6. Security Considerations
 
 It is imperative that implementations use ephemeral keys for every
 handshake as the security properties of the Kyber KEM are totally lost
@@ -294,7 +334,7 @@ software. It is hoped that the addition of a quantum resistant
 algorithm will provide forward secrecy even in the event that large
 scale quantum computers are applied to historical intercepts.
 
-## 6. Acknowledgments
+## 7. Acknowledgments
 
 I would like to thank Trevor Perrin for providing feedback during the
 design of this protocol, and answering questions regarding Noise.
@@ -321,11 +361,11 @@ year = {2017}
 }
 ```
 
-**KYBER**
+**XWING**
 
-Joppe Bos, Léo Ducas, Eike Kiltz, Tancrède Lepoint, Vadim Lyubashevsky, John M. Schanck, Peter Schwabe, Gregor Seiler, Damien Stehlé
-"CRYSTALS -- Kyber: a CCA-secure module-lattice-based KEM",
-https://cryptojedi.org/papers/kyber-20180716.pdf
+Manuel Barbosa, Deirdre Connolly, João Diogo Duarte, Aaron Kaiser, Peter Schwabe, Karoline Varner, Bas Westerbaan
+"X-Wing: The Hybrid KEM You’ve Been Looking For",
+https://eprint.iacr.org/2024/039.pdf
 
 **NOISE**
 


client2/send.go
```diff
@@ -0,0 +1,162 @@
...
+	rng := rand.NewMath()
+	p, t, err := path.New(rng, c.cfg.SphinxGeometry, doc, recipient, src, dst, surbID, baseTime, true, isForward)
+	if err == nil {
+		c.logPath(doc, p)
+	}
+
+	return p, t, err
+}
+
+func (c *Client) logPath(doc *cpki.Document, p []*path.PathHop) error {
```
threebithacker, line 152: instead duplicating this from `decoy` we should probably have it be a method of `path`, taking a logger as argument?

client2/send.go
```diff
@@ -0,0 +1,162 @@
...
+	s, err := path.ToString(doc, p)
+	if err != nil {
+		return err
```
threebithacker, line 155: Seems prudent to log this error too.

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+	cpki "github.com/katzenpost/katzenpost/core/pki"
+	sConstants "github.com/katzenpost/katzenpost/core/sphinx/constants"
+	"github.com/katzenpost/katzenpost/core/worker"
+)
+
+// ThinResponse is used to encapsulate a message response
+// that are passed to the client application.
+type ThinResponse struct {
+
+	// SURBID, a unique indentifier for this response,
```
threebithacker, line 26:
```suggestion
	// SURBID, a unique identifier for this response,
```

client2/send.go
```diff
@@ -0,0 +1,162 @@
...
+			panic("source provider cannot be nil")
+		}
+
+		fwdPath, then, err := c.makePath(request.RecipientQueueID, request.DestinationIdHash, request.SURBID, now, true)
+		if err != nil {
+			return nil, nil, 0, err
+		}
+
+		revPath := make([]*path.PathHop, 0)
+		if request.SURBID != nil {
```
threebithacker, line 54: if the SURBID is not `nil`, won't it expire if we hit a wrap-around? (aren't SURBs tied to epoch?)
if so, that's another argument for relying on `epoch` as an argument to the function or property of `request` rather than `time.Now()` after a bunch of operations.

client2/send.go
```diff
@@ -0,0 +1,162 @@
...
+		if time.Since(start) > budget {
+			continue
+		}
+
+		// It is possible, but unlikely that a series of delays exceeding
+		// the PKI publication imposted limitations will be selected.  When
+		// that happens, the path selection must be redone.
+		if then.Sub(now) < epochtime.Period*2 {
+			if request.WithSURB {
+				payload := make([]byte, 2, 2+c.geo.SURBLength+len(request.Payload))
```
threebithacker, line 75: what is the magic `2` here?

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+func (t *ThinClient) Close() error {
+	err := t.unixConn.Close()
+	t.Halt()
+	return err
+}
+
+// Dial dials the client daemon via our agreed upon abstract unix domain socket.
+func (t *ThinClient) Dial() error {
+	t.log.Debug("Dial begin")
+	srcUnixAddr, err := net.ResolveUnixAddr("unixpacket", "@katzenpost_golang_thin_client")
```
threebithacker, line 78: how does the `srcUnixAddr` work? can we run more than one client connecting to `@katzenpost` with the same src addr? (never seen src addr for unix sockets before, curious about the semantics)

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	// WAIT UNTIL we have a Noise cryptographic connection with an edge node
+	t.log.Debugf("Waiting for a connection status message")
+	message1, err := t.readNextMessage()
+	if err != nil {
+		return err
+	}
+	if !message1.IsStatus {
+		panic("did not receive a connection status message")
```
threebithacker, line 101:
```suggestion
		panic("did not receive a connection status message as the first message on fresh thin client connection")
```
Or something like that; the important part is that we expect `IsStatus` as the very first thing. We should probably log what the hell else we got instead.

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	if !message1.IsConnected {
+		return errors.New("not connected")
```
threebithacker, line 104: little bit more context here would probably be good. But I think it should be OK to launch a connection to a katzenpost daemon that is currently not connected. Best case it will connect and then you will get a new status message immediately; worst case it will never connect and then you'll have wasted a lot of time establishing and tearing down unix sockets.
If we just stay as subscribers, we also don't have to implement backoffs or delays.

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+	for {
+		select {
+		case <-t.HaltCh():
+			return
+		default:
+		}
+
+		message, err := t.readNextMessage()
+		if err != nil {
+			t.log.Infof("thin client ReceiveMessage failed: %v", err)
```
threebithacker, line 128:
```suggestion
			t.log.Errorf("thin client ReceiveMessage failed: %v", err)
```
we want to show this when spam is turned off, I think. It usually implies that the katzenpost daemon has crashed ?

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+	if !message1.IsConnected {
+		return errors.New("not connected")
+	}
+
+	t.log.Debugf("Waiting for a PKI doc message")
+	message2, err := t.readNextMessage()
+	if err != nil {
+		return err
+	}
+	t.parsePKIDoc(message2.Payload)
```
threebithacker, line 112: this can fail

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+			return
+		}
+
+		t.log.Debug("THIN CLIENT WORKER RECEIVED A MESSAGE")
+
+		switch {
+		case message.IsStatus == true:
+			t.isConnected = message.IsConnected
+		case message.IsPKIDoc == true:
+			t.parsePKIDoc(message.Payload)
```
threebithacker, line 140: this can fail.

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+			if message.Payload == nil {
+				t.log.Infof("message.Payload is nil")
+			}
+			response := ThinResponse{
+				SURBID:  message.SURBID,
+				ID:      message.ID,
+				Payload: message.Payload,
+			}
+			select {
+			case <-t.HaltCh():
```
threebithacker, line 151: should  we print to debug log that we are shutting down the worker? (same in `select {` above)

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+	blob, err := cbor.Marshal(req)
+	if err != nil {
+		return err
+	}
+	count, _, err := t.unixConn.WriteMsgUnix(blob, nil, t.destUnixAddr)
+	if err != nil {
+		return err
+	}
+	if count != len(blob) {
+		return fmt.Errorf("SendMessage error: wrote %d instead of %d bytes", count, len(blob))
```
threebithacker, line 199: Maybe this should be a panic; this really shouldn't happen (non-nil err, `count!=len(blob)`) ? I'd expect a `err` at least.

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	return nil
+}
+
+// SendMessage takes a message payload, a destination node, destination queue ID and a SURB ID and sends a message
+// along with a SURB so that you can later receive the reply along with the SURBID you choose.
+// This method of sending messages should be considered to be asynchronous because it does NOT actually wait until
+// the client daemon sends the message. Nor does it wait for a reply. The only blocking aspect to it's behavior is
+// merely blocking until the client daemon receives our request to send a message.
```
threebithacker, line 209: Good docstring, needs a small update to account for `destNode` / `destQueue` being two different things.
Who makes the SURB ID? Is the client supposed to make it up, or is a reply to something else?
(If the client should make it up, we should explain that)

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+// ReceiveMessage blocks until a message is received.
+// Use ResponseChan instead if you want an async way to receive messages.
+func (t *ThinClient) ReceiveMessage() (*[sConstants.SURBIDLength]byte, []byte) {
+	resp := <-t.receivedCh
+	return resp.SURBID, resp.Payload
+}
+
+func (t *ThinClient) ARQSend(id *[MessageIDLength]byte, payload []byte, destNode *[32]byte, destQueue []byte) error {
```
threebithacker, line 267: maybe rename `id` to `requestId` or something slightly more descriptive?
This is public but it's missing a doc string and it's not clear who should call this and when.

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+			Prefix: "thin_client",
+			Level:  log.DebugLevel,
+		}),
+		receivedCh: make(chan ThinResponse),
+	}
+}
+
+// Close halts the thin client worker thread and closes the socket
+// connection with the client daemon.
+func (t *ThinClient) Close() error {
```
threebithacker, line 69: not sure but should this also close the `receivedCh` to signal it to downstream readers?

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+
+// ThinClient is the client that handles communication between the mixnet application
+// and the client daemon. It does not do any encryption or decryption or checking
+// of cryptographic signatures; those responsibilities are left to the client daemon
+// process.
+type ThinClient struct {
+	worker.Worker
+
+	log          *log.Logger
+	unixConn     *net.UnixConn
```
threebithacker, line 45: probably need a lock protecting unixConn too if we want to support repeated/concurrent calls to `Dial()` and `Close()`.

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+// ThinResponse is used to encapsulate a message response
+// that are passed to the client application.
+type ThinResponse struct {
+
+	// SURBID, a unique indentifier for this response,
+	// which should precisely match the application's chosen
+	// SURBID of the sent message.
+	SURBID *[sConstants.SURBIDLength]byte
+
+	ID *[MessageIDLength]byte
```
threebithacker, line 31: what is this `ID` and who makes them?

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+// ARQReceiveMessage blocks until a message is received.
+// Use ResponseChan instead if you want an async way to receive messages.
+func (t *ThinClient) ARQReceiveMessage() (*[MessageIDLength]byte, []byte) {
+	resp := <-t.receivedCh
+	return resp.ID, resp.Payload
+}
+
+func (t *ThinClient) BlockingSendReliableMessage(id *[MessageIDLength]byte, payload []byte, destNode *[32]byte, destQueue []byte) (reply []byte, err error) {
```
threebithacker, line 298: this `id` thing could also use a longer name, so as not to confuse the requestID with the other IDs.
ie instead of `id` and `id2` I'd like to see `request_id` and `response_id`

client2/thin.go
```diff
@@ -0,0 +1,308 @@
...
+	err = t.ARQSend(id, payload, destNode, destQueue)
+	if err != nil {
+		return nil, err
+	}
+	id2, reply := t.ARQReceiveMessage()
+	if !bytes.Equal(id[:], id2[:]) {
+		return nil, errors.New("received unexpected ARQ reply")
```
threebithacker, line 305: we will hit this if we allow concurrent BlockingSendReliableMessage.
If we don't want that we should have more locking, and/or not expose `ARQReceiveMessage` and `ARQSend`.

Or do some complicated thing where callers to ARQSend can register callback channels.

Or a simpler thing where we register `request_id` in a map to signal that we want to buffer responses in a `map[ID] -> ThinResponse`, and then BlockingSendReliableMessage can call ARQReceiveMessage until it gets that. But then using the API yourself becomes harder. I'm not sure.

core/sphinx/path/path_factory.go
```diff
@@ -0,0 +1,46 @@
...
+// ComposePath is used to compose a Sphinx packet path. Returns
+// path and round trip time or an error.
+func (d *PathFactory) ComposePath(
+	geo *geo.Geometry,
+	doc *pki.Document,
+	srcMix *[32]byte,
+	dstId []byte,
+	dstMix *[32]byte,
+	surbID *[constants.SURBIDLength]byte,
+	baseTime time.Time,
```
threebithacker, line 34: does this need `time.Time` or could it be using epochs?

core/sphinx/packet_factory/packet_factory.go
```diff
@@ -0,0 +1,136 @@
...
+	"time"
+
+	"github.com/katzenpost/katzenpost/core/crypto/rand"
+	"github.com/katzenpost/katzenpost/core/epochtime"
+	"github.com/katzenpost/katzenpost/core/pki"
+	"github.com/katzenpost/katzenpost/core/sphinx"
+
+	"github.com/katzenpost/katzenpost/core/sphinx/constants"
+	"github.com/katzenpost/katzenpost/core/sphinx/geo"
+	"github.com/katzenpost/katzenpost/core/sphinx/path"
```
threebithacker, line 14: can we import this as `sphinx_path` instead? it reads very generic when you just see `path.PathFactory`.
Like perhaps in *this* module it's fine but in other modules it becomes confusing, I think.

core/sphinx/packet_factory/packet_factory.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (p *PacketFactory) ComposePacket(
+	doc *pki.Document,
+	srcId []byte,
+	srcMix *[32]byte,
+	dstId []byte,
+	dstMix *[32]byte,
+	surbID *[constants.SURBIDLength]byte,
+	message []byte) (packet []byte, surbKey []byte, rtt time.Duration, err error) {
```
threebithacker, line 52: There seems to be some code cloning between this and ComposeSphinxPacket in `minclient/send.go` ?

core/sphinx/kemsphinx_test.go
```diff
@@ -124,34 +127,34 @@ func newKEMPathVector(require *require.Assertions, mykem kem.Scheme, nrHops int,
 	}
 
 	// Assemble the path vector.
-	path := make([]*PathHop, nrHops)
-	for i := range path {
-		path[i] = new(PathHop)
-		copy(path[i].ID[:], nodes[i].id[:])
-		path[i].KEMPublicKey = nodes[i].publicKey
+	p := make([]*path.PathHop, nrHops)
```
threebithacker, line 130: is the rename to `p` strictly necessary here ? it kind of looks like it's to avoid the clash with the module name `path`, which again hints to me that we maybe should find a different name for that module.

core/sphinx/kemsphinx_benchmark_test.go
```diff
@@ -103,24 +104,24 @@ func newBenchKEMPathVector(mykem kem.Scheme, nrHops int, isSURB bool) ([]*kemNod
 	}
 
 	// Assemble the path vector.
-	path := make([]*PathHop, nrHops)
-	for i := range path {
-		path[i] = new(PathHop)
-		copy(path[i].ID[:], nodes[i].id[:])
-		path[i].KEMPublicKey = nodes[i].publicKey
+	p := make([]*path.PathHop, nrHops)
```
threebithacker, line 107: same here re: `path` module name.

core/pki/document_test.go
```diff
@@ -181,4 +181,13 @@ func TestDocument(t *testing.T) {
 		require.NoError(err)
 		require.True(bytes.Equal(d, d2))
 	}
+
+	// test that we can unmarshal a document stripped of signatures
```
threebithacker, line 185: we could maybe also take the opportunity to test that `StripSignatures()` lives up to its name here, and additionally that the **verifying** deserializer can no longer read it.

core/pki/descriptor.go
```diff
@@ -152,11 +174,9 @@ func (d *MixDescriptor) UnmarshalBinary(data []byte) error {
 	if err != nil {
 		return err
 	}
-	_, err = cert.Verify(d.IdentityKey, data)
-	if err != nil {
-		return err
+	if sigs != nil && len(sigs) != 0 {
+		d.Signature = &sigs[0]
```
threebithacker, line 178: len(sigs) > 1 used to return `ErrTooManySignatures` I think? now we we don't care?

client2/timer_queue.go
```diff
@@ -0,0 +1,76 @@
...
+	sConstants "github.com/katzenpost/katzenpost/core/sphinx/constants"
+	"github.com/katzenpost/katzenpost/core/worker"
+)
+
+type TimerQueue struct {
+	worker.Worker
+
+	queue  *queue.PriorityQueue
+	timer  *time.Timer
+	mutex  sync.Mutex
```
threebithacker, line 19: maybe this could be an RWLock to enable non-blocking `t.Len()`  and `t.queue.Peek()` ?

client2/poisson_process_test.go
```diff
@@ -0,0 +1,42 @@
+// SPDX-FileCopyrightText: © 2023 David Stainton
+// SPDX-License-Identifier: AGPL-3.0-only
+
```
threebithacker, line 3: - What are we testing in this file?
  - What do the operations do for test?
  - What is the expected outcome for each test?
  - Why is that outcome expected
documenting this lets us modify this code in the future without spending ages reverse-engineering the logic.

threebithacker (changes requested):

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// ARQMessage is used by ARQ.
+type ARQMessage struct {
+
+	// AppID identifies the application sending/receiving the message/reply.
+	AppID uint64
+
+	// ID is the unique message identifier
+	ID *[MessageIDLength]byte
```
threebithacker, line 38:
```suggestion
	MessageID *[MessageIDLength]byte
```
it's governed by `MessageID` length, and then we are not left at callsites wondering what kind of `id` something is

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+	RecipientQueueID []byte
+
+	// Payload is the message payload
+	Payload []byte
+
+	// SURBID is the SURB identifier.
+	SURBID *[sConstants.SURBIDLength]byte
+
+	// Key is the SURB decryption keys
+	Key []byte
```
threebithacker, line 54:
```suggestion
	SURBDecryptionKey []byte
```
might as well, less confusion that way

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+	// RecipientQueueID is the queue identity which will receive the message.
+	RecipientQueueID []byte
+
+	// Payload is the message payload
+	Payload []byte
+
+	// SURBID is the SURB identifier.
+	SURBID *[sConstants.SURBIDLength]byte
+
+	// Key is the SURB decryption keys
```
threebithacker, line 53:
```suggestion
	// Key is the SURB decryption key
```

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// ID is the unique message identifier
+	ID *[MessageIDLength]byte
+
+	// DestinationIdHash is 32 byte hash of the destination Provider's
+	// identity public key.
+	DestinationIdHash *[32]byte
+
+	// RecipientQueueID is the queue identity which will receive the message.
+	RecipientQueueID []byte
```
threebithacker, line 45: Is this free-form or should it also be an array like `[sphinx.constants.RecipientIDLength]`?

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// AppID identifies the application sending/receiving the message/reply.
+	AppID uint64
+
+	// ID is the unique message identifier
+	ID *[MessageIDLength]byte
+
+	// DestinationIdHash is 32 byte hash of the destination Provider's
+	// identity public key.
+	DestinationIdHash *[32]byte
```
threebithacker, line 42: this looks like it's the same 32 byte DestinationIdHash that's used in `commands.go`. Consider also adding a constant for the size of that.

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+
+	"github.com/charmbracelet/log"
+
+	"github.com/katzenpost/katzenpost/core/crypto/rand"
+	sConstants "github.com/katzenpost/katzenpost/core/sphinx/constants"
+)
+
+const (
+	// MessageIDLength is the length of a message ID in bytes.
+	MessageIDLength = 16
```
threebithacker, line 19: Who uses this message ID, and who generates it? Is it unique to the ARQ stuff or to the client2  stack or is it a deeper concept?

client2/commands.go
```diff
@@ -0,0 +1,80 @@
...
+	// in the Sphinx payload.
+	WithSURB bool
+
+	// SURBID must be a unique identity for each request.
+	// This field should be nil if WithSURB is false.
+	SURBID *[sConstants.SURBIDLength]byte `cbor:surbid`
+
+	// AppID must be a unique identity for the client application
+	// that is sending this Request.
+	AppID uint64 `cbor:app_id`
```
threebithacker, line 48: if it has to be unique we might as well make it a 128 bit byte array, then we don't have to worry about the threat of local applications bruteforcing each other's IDs.

client2/config/config.go
```diff
@@ -0,0 +1,341 @@
...
+		return err
+	}
+
+	m := data["Addresses"].(map[string]interface{})
+	p.Addresses = make(map[string][]string)
+
+	for k, v := range m {
+		values := make([]string, 0)
+		if v == nil {
+			return fmt.Errorf("error: KEY %s has nil value\n", k)
```
threebithacker, line 176: `k` here is inside `Addresses` inside the provider `p` ?
```suggestion
			return fmt.Errorf("error: KEY Providers.%s.[Addresses].%s has nil value\n", p.Name, k)
```

client2/config/config.go
```diff
@@ -0,0 +1,341 @@
...
+		} else {
+			vals := v.([]interface{})
```
threebithacker, line 178: should it be an error if we end up with an empty values here? can that happen?

client2/config/config.go
```diff
@@ -0,0 +1,341 @@
...
+func (lCfg *Logging) validate() error {
+	lvl := strings.ToUpper(lCfg.Level)
+	switch lvl {
+	case "ERROR", "WARNING", "INFO", "DEBUG":
+	case "":
+		lCfg.Level = defaultLogLevel
+	default:
+		return fmt.Errorf("config: Logging: Level '%v' is invalid", lCfg.Level)
+	}
+	lCfg.Level = lvl // Force uppercase.
```
threebithacker, line 63:
```suggestion
	lCfg.Level = lvl

File: client2/config/config.go [hunk 40]
```diff
@@ -0,0 +1,341 @@
...
+
+	// File specifies the log file, if omitted stdout will be used.
+	File string
+
+	// Level specifies the log level.
+	Level string
+}
+
+func (lCfg *Logging) validate() error {
+	lvl := strings.ToUpper(lCfg.Level)
```
threebithacker, line 55:
```suggestion
	lvl := strings.ToUpper(lCfg.Level) // Force uppercase.
```

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+		message.Key = k
+		message.ReplyETA = rtt
+		message.SentAt = time.Now()
+		message.Retransmissions += 1
+
+		a.surbIDMap[*newsurbID] = message
+		priority := uint64(message.SentAt.Add(message.ReplyETA).Add(RoundTripTimeSlop).UnixNano())
+		a.timerQueue.Push(priority, surbID)
+
+		err = a.sphinxComposerSender.SendSphinxPacket(pkt)
```
threebithacker, line 151: does the lock need to be help for this to work?

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+// HandleACK removes the map entry for the given SURB ID AND returns
+// the APP ID and SURB Key so that the reply and be decrypted and routed
+// to the correct application.
+func (a *ARQ) HandleAck(surbID *[sConstants.SURBIDLength]byte) (*replyDescriptor, error) {
+	a.lock.Lock()
+
+	m, ok := a.surbIDMap[*surbID]
+	if !ok {
+		a.log.Error("failed to find SURB ID in ARQ map")
+		return nil, errors.New("failed to find SURB ID in ARQ map")
```
threebithacker, line 181: need to relinquish the lock here or we deadlock.
probably easiest to just `defer a.lock.Unlock()` at the top here

client2/arq.go
```diff
@@ -0,0 +1,241 @@
...
+		AppID:    appid,
+		ID:       id,
+		SURBID:   surbID,
+		Payload:  payload,
+		Key:      k,
+		SentAt:   time.Now(),
+		ReplyETA: rtt,
+	}
+	a.surbIDMap[*surbID] = message
+	p := time.Duration(message.ReplyETA + RoundTripTimeSlop)
```
threebithacker, line 229: beyond here it doesn't look like we need the lock anymore?

threebithacker (changes requested):
ARQ locking stuff

authority/voting/client/client.go
```diff
@@ -201,12 +202,12 @@ func (p *connector) allPeersRoundTrip(ctx context.Context, linkKey wire.PrivateK
 	for _, peer := range p.cfg.Authorities {
 		conn, err := p.initSession(ctx, doneCh, linkKey, signingKey, peer)
 		if err != nil {
-			p.log.Noticef("pki/voting/client: failure to connect to Authority %s (%x)\n", peer.Identifier, peer.IdentityPublicKey.Sum256())
+			p.log.Infof("pki/voting/client: failure to connect to Authority %s (%x)\n", peer.Identifier, peer.IdentityPublicKey.Sum256())
```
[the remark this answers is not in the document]
david415: no, it's just a log message. it wasn't meant to specify the precise module name.

client2/send.go
```diff
@@ -0,0 +1,162 @@
...
+
+// SendSphinxPacket sends the given Sphinx packet.
+func (c *Client) SendSphinxPacket(pkt []byte) error {
+	return c.conn.sendPacket(pkt)
+}
+
+// ComposeSphinxPacket is used to compose Sphinx packets.
+func (c *Client) ComposeSphinxPacket(request *Request) ([]byte, []byte, time.Duration, error) {
+
+	if len(request.RecipientQueueID) > sConstants.RecipientIDLength {
```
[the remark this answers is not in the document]
david415: no. here i'm doing type checking because the field is not bounded by the name type since i used []byte for the RecipientQueueID

client2/send.go
```diff
@@ -0,0 +1,162 @@
...
+	sConstants "github.com/katzenpost/katzenpost/core/sphinx/constants"
+	"github.com/katzenpost/katzenpost/core/sphinx/path"
+)
+
+// SendSphinxPacket sends the given Sphinx packet.
+func (c *Client) SendSphinxPacket(pkt []byte) error {
+	return c.conn.sendPacket(pkt)
+}
+
+// ComposeSphinxPacket is used to compose Sphinx packets.
```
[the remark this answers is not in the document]
david415: fixed.

client2/send.go
```diff
@@ -0,0 +1,162 @@
...
+func (c *Client) ComposeSphinxPacket(request *Request) ([]byte, []byte, time.Duration, error) {
```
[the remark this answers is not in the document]
david415: yes, i hear what you are saying. i'll take a look at doing that soon.

client2/thin.go
david415: fixed

client2/thin.go
david415: no. clients are expected to launch multiple thin clients not just one!

client2/thin.go
david415: it's a message id. the client application choses these IDs

client2/thin.go
david415: same id

client2/thin.go
david415: there is no request id. there are only message IDs and SURB IDs

authority/voting/server/state.go
```diff
@@ -1875,7 +1875,7 @@ func (s *state) backgroundFetchConsensus(epoch uint64) {
 		go func() {
 			cfg := &client.Config{
 				LinkKey:       s.s.linkKey,
-				LogBackend:    s.s.logBackend,
+				LogBackend:    s.s.logBackend.GetLogWriter("pki", "Info"),
```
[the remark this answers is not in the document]
david415: pki is the context for the logger

ping/voting_docker_mixnet.toml
```diff
@@ -1,130 +0,0 @@
-
-DataDir = "./testdata/"
-
-[UpstreamProxy]
```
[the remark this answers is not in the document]
david415: docker

core/sphinx/path/path_factory.go
david415: path select uses time

core/sphinx/packet_factory/packet_factory.go
david415: minclient is going away forever.

core/sphinx/kemsphinx_benchmark_test.go
david415: calm down

core/pki/descriptor.go
david415: unmarshal binary should be used for cryptography. all wrong.

client2/timer_queue.go
david415: fixed

client2/arq.go
david415: it's the message id!

core/sphinx/packet_factory/packet_factory.go
david415: fixed

core/pki/document_test.go
david415: no. those changes were made a long time ago. we can mess with it in another PR.

core/pki/descriptor.go
david415: i'm going to have to fix this bullshit later.

client2/poisson_process_test.go
david415: ive added commentary to each test

client2/arq.go
david415: fixed

client2/arq.go
david415: fixed

client2/arq.go
david415: at this time we don't have such constant but maybe we should

client2/config/config.go
david415: fixed

client2/config/config.go
david415: fixed

client2/arq.go
david415: yes

client2/arq.go
david415: fixed

client2/arq.go
david415: fixed

PR #209, opened 2023-01-14
Use priority queue for decoys instead of AVL tree

david415:
DO NOT MERGE

### DIFF
--- a/server/internal/decoy/decoy.go
+++ b/server/internal/decoy/decoy.go
@@ -28,10 +28,13 @@ import (
 	"sync"
 	"time"
 
+	"gopkg.in/op/go-logging.v1"
+
 	"github.com/katzenpost/katzenpost/core/crypto/rand"
 	"github.com/katzenpost/katzenpost/core/epochtime"
 	"github.com/katzenpost/katzenpost/core/monotime"
 	"github.com/katzenpost/katzenpost/core/pki"
+	"github.com/katzenpost/katzenpost/core/queue"
 	"github.com/katzenpost/katzenpost/core/sphinx"
 	"github.com/katzenpost/katzenpost/core/sphinx/commands"
 	sConstants "github.com/katzenpost/katzenpost/core/sphinx/constants"
@@ -42,8 +45,6 @@ import (
 	"github.com/katzenpost/katzenpost/server/internal/packet"
 	"github.com/katzenpost/katzenpost/server/internal/pkicache"
 	"github.com/katzenpost/katzenpost/server/internal/provider/kaetzchen"
-	"gitlab.com/yawning/avl.git"
-	"gopkg.in/op/go-logging.v1"
 )
 
 const maxAttempts = 3
@@ -51,11 +52,10 @@ const maxAttempts = 3
 var errMaxAttempts = errors.New("decoy: max path selection attempts exceeded")
 
 type surbCtx struct {
-	id      uint64
-	eta     time.Duration
-	sprpKey []byte
-
-	etaNode *avl.Node
+	id       uint64
+	eta      time.Duration
+	sprpKey  []byte
+	priority uint64
 }
 
 type decoy struct {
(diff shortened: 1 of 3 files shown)


server/internal/decoy/decoy.go
```diff
@@ -72,11 +72,16 @@ type decoy struct {
...
 
-	surbETAs   *avl.Tree
+	surbETAs   *queue.PriorityQueue
 	surbStore  map[uint64]*surbCtx
 	surbIDBase uint64
 }
 
+func addPriority(ctx *surbCtx) {
+	now := monotime.Now()
+	ctx.priority = uint64(now + ctx.eta)
```
mixmasala, line 82: can we ensure that priority is always unique?

server/internal/decoy/decoy.go
```diff
@@ -385,30 +379,7 @@ func (d *decoy) loadAndDeleteSURBCtx(id uint64) *surbCtx {
...
-				return ctx
-			}
-		}
-		panic("BUG: SURB missing from node ETA list")
-	}
-
-	d.surbETAs.Remove(ctx.etaNode)
-	ctx.etaNode = nil
-
+	d.surbETAs.RemovePriority(ctx.priority)
```
mixmasala, line 382: check return value of RemovePriority != nil ( see panic("BUG.."))

mixmasala (changes requested):
I think this is probabalistically safe as our PriorityQueue may not handle items with the same priority appropriately; it would also be nice to see a comparison benchmark between the implementations.

server/internal/decoy/decoy.go
david415: we don't need to ensure it's unique

server/internal/decoy/decoy.go
```diff
@@ -365,14 +365,8 @@ func (d *decoy) storeSURBCtx(ctx *surbCtx) {
...
-	ctxList := []*surbCtx{ctx}
-	ctx.etaNode = d.surbETAs.Insert(ctxList)
-	if nCtxList := ctx.etaNode.Value.([]*surbCtx); nCtxList[0] != ctx {
-		// Existing node, append the context to the list stored
-		// as the node value.  This should be exceedingly rare.
-		ctx.etaNode.Value = append(nCtxList, ctx)
-	}
-
+	addPriority(ctx)
+	d.surbETAs.Enqueue(ctx.priority, ctx)
```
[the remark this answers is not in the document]
david415: no it doesn't loose the reference. i just now wrote a test for this.

server/internal/decoy/decoy.go
david415: check the return value for what!?

david415: this cannot crash. try it.

server/internal/decoy/decoy.go
```diff
@@ -423,29 +394,19 @@ func (d *decoy) sweepSURBCtxs() {
...
 			break
 		}
-
-		for _, ctx := range surbCtxs {
-			delete(d.surbStore, ctx.id)
-			// TODO: At some point, this should do more than just log.
-			d.log.Debugf("Sweep: Lost SURB ID: 0x%08x ETA: %v (DeltaT: %v)", ctx.id, ctx.eta, now-ctx.eta)
-			swept++
+		ctx := entry.Value.(*surbCtx)
+		now := monotime.Now()
```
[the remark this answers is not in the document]
david415: fixed

david415:
> I think this is probabalistically safe as our PriorityQueue may not handle items with the same priority appropriately; it would also be nice to see a comparison benchmark between the implementations.

i think that is an incorrect assumption

david415:
terrible no progress not sure how to fix it using the heap library.

mixmasala:
> > I think this is probabalistically safe as our PriorityQueue may not handle items with the same priority appropriately; it would also be nice to see a comparison benchmark between the implementations.
> 
> i think that is an incorrect assumption

could I help clarify?

Issue #213, opened 2023-01-16
fix mixserver memory queue


david415:
The priority queue used by our core/queue implementation of a priority queue does not properly handle insertions of duplicate priorities. the old one gets overwritten. This is causing subtle bugs and probably packet loss.

This code snippet shows that we don't do anything to prevent duplicate priorities since they are chosen with entropy:
https://github.com/katzenpost/katzenpost/blob/main/server/internal/scheduler/queue_mem.go#L57-L78

mixmasala:
merged in #216

Issue #212, opened 2023-01-15
replace core/queue ASAP!!!


david415:
NOTE: we use core/queue, our priority queue implementation based on golang's heap package. core/queue is used by every component of katzenpost.

our priority queue does not properly handle duplicate priorities. this can lead to subtle bugs.

possible solutions:

1. write a well tested priority queue from scratch
2. find an existing priority queue in golang, that has the properties we want

Whatever we replace it with, it must be a datastructure that has:

1. O(log n) insertion of entry by priority
2. O(1) pop/deletion of latest priority
3. O(log n) deletion by specified priority
4. it must handle duplicate priorities properly and not loose any data!

mixmasala:
I think we agree that we do not need the ability to remove by priority - I wrote this thinking I needed a mechanism to remove messages from a PriorityQueue that held Messages until they were ACK'd, where the soonest-to-timeout Message is at the front of the PriorityQueue

However, it is much simpler/cheaper to just let the timer fire for each message, and check whether it was ACK'd from a different datastructure. Amusingly, I already did that, so the fix was to just remove calls to RemovePriority.

mixmasala:
Fix merged in #216

PR #214, opened 2023-01-16
Fix priority queue

david415:
This is a fix to the mixserver. This fixes:
https://github.com/katzenpost/katzenpost/issues/213

Also fixes:
https://github.com/katzenpost/katzenpost/issues/212

except that it doesn't allow for O(log n) deletion by priority... which currently is only needed by our buggy "client" package.

### DIFF
--- a/client/corequeue/priority_queue.go
+++ b/client/corequeue/priority_queue.go
@@ -1,0 +1,157 @@
+// priority_queue.go - Min-Heap based priority queue.
+// Copyright (C) 2017, 2018  David Anthony Stainton, Yawning Angel
+//
+// This was inspired by the priority queue example in the godocs:
+// https://golang.org/pkg/container/heap/
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Package queue implements a priority queue.
+package queue
+
+import (
+	"container/heap"
+	"math/rand"
+)
+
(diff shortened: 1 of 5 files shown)


david415:
> I do not agree that this is the right fix; rather that we should either update PriorityQueue to support adding entries with the same Priority or define the behavior of PriorityQueue to not support multiple items with the same Priority, and to return an error in that case. I think that the frequency this occurs is likely much lower than base packetloss rate and a nonissue. @Yawning might have an opinion we can consider as I suspect they considered this edge case when they wrote the PriorityQueue impl

I wrote the priority queue... not yawning.

mixmasala:
> I do not agree that this is the right fix; rather that we should either update PriorityQueue to support adding entries with the same Priority or define the behavior of PriorityQueue to not support multiple items with the same Priority, and to return an error in that case. I think that the frequency this occurs is likely much lower than base packetloss rate and a nonissue. @Yawning might have an opinion we can consider as I suspect they considered this edge case when they wrote the PriorityQueue impl

per conversation with dawuud I'm more convinced that he is right and we should get rid of the extra complexity I added:

```
(because the mixserver should never remove a packet from the queue by a specific priority)
there is another way to fix the client as well
which is to just keep track of ACK'd messages
in a seperate datastructure, so that you dont need to remove them from the PriorityQueue
the thing I built that client uses is a single timer that is updated with the PriorityQueue, but if the timer fires and the thing is already ACK'd it's kinda OK to just check that it was ACK'd and not restransmit
and probably simplify that stuff as well
```

mixmasala:
```
katzenpost# grep RemovePriority -r .
./client/timer_queue.go:			mo := a.priq.RemovePriority(priority)
./core/queue/priority_queue_test.go:		e := q.RemovePriority(v.Priority)
./core/queue/priority_queue_test.go:		e := q.RemovePriority(v.Priority)
./core/queue/priority_queue_test.go:	e := q.RemovePriority(1234)
./core/queue/priority_queue.go:// RemovePriority removes and returns element from the heap with given priority or nil
./core/queue/priority_queue.go:func (q *PriorityQueue) RemovePriority(priority uint64) interface{} {
```
It looks like there is only the client retransmission timer that makes use of RemovePriority, so the goal I think should be to revert the PriorityQueue and make TimerQueue keep track of acknowledgements - it does mean the timer fires for every message regardless if it was ACK'd or not but this should be a performance win overall .

mixmasala:
@dawuud do you mind folding in changes I made in  #216 ?

Issue #207, opened 2023-01-14
decoy.go:438 panic: runtime error: invalid memory address or nil pointer dereference


leif:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x859cf6]

goroutine 41 [running]:
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).sweepSURBCtxs(0xc000000140)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:438 +0x2b6
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).worker(0xc000000140)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:190 +0x53a
github.com/katzenpost/katzenpost/core/worker.(*Worker).Go.func1()
        /home/katzenpost/katzenpost/core/worker/worker.go:39 +0x5a
created by github.com/katzenpost/katzenpost/core/worker.(*Worker).Go
        /home/katzenpost/katzenpost/core/worker/worker.go:37 +0xb9
panic: interface conversion: interface {} is *decoy.surbCtx, not []*decoy.surbCtx
```

this looks related to the two avl problems in #133, and also to another crash which i'll file after this one.

Issue #208, opened 2023-01-14
decoy.go:465 panic: interface conversion: interface {} is *decoy.surbCtx, not []*decoy.surbCtx


leif:
```
panic: interface conversion: interface {} is *decoy.surbCtx, not []*decoy.surbCtx

goroutine 41 [running]:
github.com/katzenpost/katzenpost/server/internal/decoy.New.func1({0x936f40?, 0xc00afb9bd8?}, {0x92fc20?, 0x0?})
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:465 +0xc5
gitlab.com/yawning/avl%2egit.(*Tree).Insert(0xc0000126c0, {0x936f40?, 0xc00afb9bd8})
        /go/pkg/mod/gitlab.com/yawning/avl.git@v0.0.0-20180224045358-04c7c776e391/avl.go:211 +0x87
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).storeSURBCtx(0xc0000003c0, 0xc00b1af740)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:369 +0xe8
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).sendLoopPacket(0xc0000003c0, 0xc0070ef5e8?, {0xc0001c5de8, 0x5, 0x20}, 0xe0bf00?, 0x46976455c6?)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:283 +0x57b
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).sendDecoyPacket(0xc0000003c0, 0xc0001c5ebc?)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:238 +0x1cd
github.com/katzenpost/katzenpost/server/internal/decoy.(*decoy).worker(0xc0000003c0)
        /home/katzenpost/katzenpost/server/internal/decoy/decoy.go:173 +0x405
github.com/katzenpost/katzenpost/core/worker.(*Worker).Go.func1()
        /home/katzenpost/katzenpost/core/worker/worker.go:39 +0x5a
created by github.com/katzenpost/katzenpost/core/worker.(*Worker).Go
        /home/katzenpost/katzenpost/core/worker/worker.go:37 +0xb9
```

leif:
(this is related to #133 and #207)

PR #215, opened 2023-01-16
Attempt to fix server decoy

david415:
fix some of the server decoy traffic bugs

### DIFF
--- a/server/internal/decoy/decoy.go
+++ b/server/internal/decoy/decoy.go
@@ -75,7 +75,6 @@ type decoy struct {
 	surbETAs   *avl.Tree
 	surbStore  map[uint64]*surbCtx
 	surbIDBase uint64
-	uniqueId   uint64
 }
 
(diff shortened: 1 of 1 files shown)


server/internal/decoy/decoy.go
```diff
@@ -75,6 +75,7 @@ type decoy struct {
 	surbETAs   *avl.Tree
 	surbStore  map[uint64]*surbCtx
 	surbIDBase uint64
+	uniqueId   uint64
```
threebithacker, line 78:
```suggestion
```
We ended up just using the 8 random bytes from the SurbID instead, so we didn't need the sequential counter after all.

threebithacker (changes requested):

threebithacker (approved):

leif:
@david415 @threebithacker do you think this fixes #207 and #208 ?

PR #216, opened 2023-01-16
Fix priority queue.1

mixmasala:
This updates client to no longer need the RemovePriority method of PriorityQueue and the associated structures. My bad.


mixmasala:
oops i forgot to get rid of the corequeue copy but this should probably still work now...

mixmasala:
lets not actually merge this though

david415 (approved):

PR #219, opened 2023-01-17
Sphinx geo pki

david415:
Needs code review. This pull request puts the Sphinx geometry into the PKI document thus parameterizing our Sphinx geometry for all components of the katzenpost network.

The authority configuration now contains the Sphinx geometry. The assumption here is that all the dirauths agree on what the Sphinx geometry should be, so they all put the same geometry into their respective dirauth configuration files.

Clients AND mixservers both download the PKI document AND THEN get the Sphinx geometry which allows them to participate in the mixnet. Interaction with mixes over the katzenpost Noise based wire protocol requires knowing the Sphinx geometry because some of the wire protocol commands rely on knowledge of the sphinx geometry in order to get encoded and decoded properly.

However using our noise protocol to interact with dirauths does not require knowledge of the Sphinx geometry. And for that purpose I created a separate noise wire protocol constructor.

Not only can this feature be used to make variously shaped mixnets with any number of hops it is also a prerequisite to "crossover points", which is a future feature of the mix server allow it to be multi-homed on two different mixnets with two different Sphinx geometries (assuming one Sphinx can fix in the payload of the other Sphinx). This is a necessary step if you want a hybrid PQ mixnet that uses CTIDH and our classical Sphinx.


threebithacker (changes requested):
Haven't been through it all, but here are comments/questions so far

authority/voting/client/client.go
```diff
@@ -374,6 +372,9 @@ func (c *Client) Get(ctx context.Context, epoch uint64) (*pki.Document, []byte,
 		c.log.Errorf("voting/Client: Get() invalid consensus document: %s", err)
 		return nil, nil, err
 	}
+	if doc.SphinxGeometry == nil {
+		panic("received pki doc with nil SphinxGeometry")
+	}
```
threebithacker, line 377: Do we need to validate that it's a *sane* geometry (positive number of hops etc?)
Doesn't this check belong in `pki.VerifyAndParseDocument` (then it could also fail gracefully instead of panicing)?

threebithacker (changes requested):

authority/voting/server/server.go
```diff
@@ -187,6 +189,12 @@ func (s *Server) halt() {
 func New(cfg *config.Config) (*Server, error) {
 	s := new(Server)
 	s.cfg = cfg
+	s.geo = cfg.SphinxGeometry
```
threebithacker, line 192: Hmm, are there cases where we don't just want to do `s.cfg.SphinxGeometry` ? maybe `Server` doesn't need its own copy of the geometry?

threebithacker (changes requested):

README.rst
```diff
@@ -15,22 +15,45 @@ https://katzenpost.mixnetworks.org/
 Building from Go source
 =======================
 
-Katzenpost now uses hybrid post quantum with classical cryptographic protocols.
-As part of this change in our cryptographic protocols we have added the use
-of Sphincs+, the stateless hash based post quantum signature scheme.
+Katzenpost now uses hybrid post quantum with classical cryptographic
```
threebithacker, line 18: `hybrid post-quantum` feels like two adjectives, we're missing a noun?

threebithacker (changes requested):

core/crypto/ecdh/ecdh.go
```diff
@@ -199,6 +199,22 @@ func (k *PrivateKey) UnmarshalBinary(data []byte) error {
...
 
+// MarshalText is an implementation of a method on the
+// TextMarshaler interface defined in https://golang.org/pkg/encoding/
+func (k *PrivateKey) MarshalText() ([]byte, error) {
+	return []byte(base64.StdEncoding.EncodeToString(k.Bytes())), nil
+}
+
+// UnmarshalText is an implementation of a method on the
+// TextUnmarshaler interface defined in https://golang.org/pkg/encoding/
+func (k *PrivateKey) UnmarshalText(data []byte) error {
```
threebithacker, line 210: maybe it'd be useful to put in the go doc that these deal with base64 ?

core/crypto/nike/nike.go
```diff
@@ -46,9 +52,12 @@ type PublicKey interface {
 	Blind(blindingFactor []byte) error
 }
 
-// Nike is an interface encapsulating a
+// Scheme is an interface encapsulating a
 // non-interactive key exchange.
-type Nike interface {
+type Scheme interface {
```
threebithacker, line 57: Scheme seems awfully generic and hard to search for, `NikeScheme` maybe?

core/crypto/nike/hybrid/hybrid.go
```diff
@@ -0,0 +1,210 @@
...
+	raw, err := base64.StdEncoding.DecodeString(string(data))
+	if err != nil {
+		return err
+	}
+	return p.FromBytes(raw)
+}
+
+func (p *publicKey) Blind(blindingFactor []byte) error {
+	err := p.first.Blind(blindingFactor[:p.scheme.first.PublicKeySize()])
+	if err != nil {
```
threebithacker, line 161: Since we may have written stuff to `p.first` in these methods, it would perhaps be safer if we called `p.Reset()` before returning `err`, to prevent people from using partially overwritten things?

core/crypto/nike/hybrid/hybrid.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (s *scheme) DerivePublicKey(privKey nike.PrivateKey) nike.PublicKey {
+	return &publicKey{
+		scheme: s,
+		first:  privKey.(*privateKey).scheme.first.DerivePublicKey(privKey.(*privateKey).first),
+		second: privKey.(*privateKey).scheme.second.DerivePublicKey(privKey.(*privateKey).second),
+	}
+}
+
+func (s *scheme) Blind(groupMember []byte, blindingFactor []byte) (blindedGroupMember []byte) {
```
threebithacker, line 78: that `BlindingFactor` takes `[s.PrivateKeySize]byte` seems like an implementation detail, it feels like there should be a constant/method for that in the interface.
This `Blind` takes a `[]byte`, and assumes that a `scheme` uses the same code to blind both `nike.PublicKey` and `nike.PrivateKey`, which also seems like a fortunate coincidence.
I think it'd be cleaner if we had separate methods for `nike.PublicKey` and `nike.PublicKey`, then we could also typecheck things (and avoid serialization to/from bytes)

threebithacker (changes requested):

authority/voting/client/client_test.go
```diff
@@ -256,7 +263,7 @@ func (d *mockDialer) mockServer(address string, linkPrivateKey wire.PrivateKey,
 	d.waitUntilDialed(address)
 	identityHash := identityPublicKey.Sum256()
 	cfg := &wire.SessionConfig{
-		Geometry:          &sphinx.Geometry{},
+		Geometry:          &geo.Geometry{},
```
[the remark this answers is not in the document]
david415: Geometry field of SessionConfig struct would have a value of nil if not explicitly set in a declaration of SessionConfig. What is the actual suggestion to make this more explicit? It's how go works.

minclient/connection.go
```diff
@@ -360,7 +361,7 @@ func (c *connection) onTCPConn(conn net.Conn) {
 
 	// Allocate the session struct.
 	cfg := &wire.SessionConfig{
-		Geometry:          sphinx.DefaultGeometry(),
+		Geometry:          c.geo,
```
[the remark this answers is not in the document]
david415: Each component of katzenpost needs to have a Sphinx geometry, therefore it's natural to copy it around. We were doing that before, only from a hardcoded geometry.

The Packet API could be changed so that it reads a copy of the Sphinx geometry from elsewhere in the server's code.

SessionConfig needs to have the Sphinx geometry because some of the noise wire commands use the geometry to determine the payload size.

genconfig/main.go
```diff
@@ -437,6 +443,15 @@ func main() {
 	s.basePort = uint16(*basePort)
 	s.lastPort = s.basePort + 1
 
+	nrHops := *nrLayers + 2
+
+	s.sphinxGeometry = geo.GeometryFromUserForwardPayloadLength(
+		ecdh.NewEcdhNike(rand.Reader),
+		2000,
```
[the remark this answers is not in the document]
david415: The old geometry had a payload of 2048 - SURB size. Please understand that the old geometry was calculated in a way that is bad for us... this PR improves the situation by letting you specify the usable payload size. The goal was never to preserve the same payload size forever. But by having a payload size we can more directly control it makes it easier for the higher level protocols to reason about how much payload size they have.

And yes we can name a variable and assign in the usable payload size which by the way is not the same as the Forwarding Length.... which is the total payload size including the SURB.

server/server.go
```diff
@@ -311,13 +314,13 @@ func New(cfg *config.Config) (*Server, error) {
 	s.log.Noticef("Server link public key hash is: %x", linkPubKeyHash[:])
 
 	if s.cfg.Debug.GenerateOnly {
-		return nil, ErrGenerateOnly
+		return ErrGenerateOnly
```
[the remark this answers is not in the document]
david415: No. We have to actually abide by the function signature and return an error. Setting err is meaningless otherwise. I think what you are remembering here is that sometimes go functions have a signature that specifies the return type with variable names. This is not one of those cases.

server/server.go
```diff
@@ -375,21 +378,30 @@ func New(cfg *config.Config) (*Server, error) {
...
 			s.log.Errorf("Failed to initialize provider backend: %v", err)
-			return nil, err
+			return nil
 		}
 	}
 
+	s.pki.StartWorker()
+
+	s.log.Noticef("Current Sphinx Geometry is:\n[SphinxGeometry]\n%s\n",
+		goo.PKI().GetSphinxGeometry().Display())
```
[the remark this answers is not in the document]
david415: So... the PKI worker is what gathers the Sphinx Geometry in the first place! Therefore that code is correct and things are currently done in the correct order.:
```
func (p *pki) StartWorker() {
	p.Go(p.worker)
	p.blockUntilConsensus()
}

func (p *pki) blockUntilConsensus() {
	<-p.blockStartChan
}

func (p *pki) worker() {

```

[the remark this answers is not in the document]
david415: Once we have a consensus then the rest of the server's code base can access the geometry via goo.PKI().GetSphinxGeometry()

authority/voting/client/client.go
david415: Yes we should have a Validate method on Geometry to make sure it is sane. I'll add that.

authority/voting/server/server.go
david415: yeah we can move the geometry reference to the config if you think that is cleaner

core/crypto/nike/nike.go
david415: It's in the "nike" module thus making the API stutter: nike.NikeScheme.
We're following circl's example here: https://github.com/cloudflare/circl/blob/main/kem/kem.go#L33

core/crypto/nike/hybrid/hybrid.go
david415: yeah sounds good

core/crypto/nike/hybrid/hybrid.go
david415: I don't understand what coincidence you are referring to. Private keys are NOT blinded. Private keys are merely a blinding factor.

authority/voting/client/client_test.go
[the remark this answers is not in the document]
david415: There are basically only one place in the code where I would want to declar SessionConfig and have Geometry be nil and that is the other constructor for the wire protocol that is only meant to communicate with the dirauths, and thus no geometry for that is needed.

server/internal/packet/packet.go
```diff
@@ -297,16 +303,21 @@ func NewPacketFromSURB(pkt *Packet, surb, payload []byte) (*Packet, error) {
 	cmds = append(cmds, nodeDelayCmd)
 
 	// Assemble the response packet.
-	respPkt, _ := New(rawRespPkt)
+	respPkt, _ := New(rawRespPkt, geo)
 	respPkt.Geometry = pkt.Geometry
```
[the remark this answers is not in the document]
david415: probably this whole module's API needs to be rewritten... it's too fucking confusing

authority/voting/client/client.go
david415: Okay! I've added the Validate method... and call it at the appropriate places

core/crypto/ecdh/ecdh.go
david415: Well okay then do we want to change it for all our crypto modules or just this one? And... can we do so in a separate ticket/PR?

core/crypto/nike/hybrid/hybrid.go
david415: done

core/crypto/nike/hybrid/hybrid.go
david415: This is NOT an implementation detail. This code works for ALL NIKEs. All NIKEs MUST MUST perform the blinding operation with a blinding factor that is equal in length to the private key... and in fact it must be a valid private key for that NIKE scheme.

Issue #56, opened 2022-11-04
DialContextFn leaks linkkey / linkkey hash to socks proxy as socks username


threebithacker:
`client/config/config.go` presently sets the `DialContextFn` like this:
```go
cfg := &nvClient.Config{
 DialContextFn:        pCfg.ToDialContext(fmt.Sprintf("nonvoting: %x", linkHash[:])),
```
or
```go
cfg := &nvClient.Config{
  DialContextFn:    pCfg.ToDialContext("nonvoting:" + nvACfg.PublicKey.String()),
```
This leaks the link key public key (or its hash) as socks username (used to make Tor trigger its stream isolation feature based on socks username).

It seems to me there might be a privacy issue luring there, and that we could just fill it with random bytes instead; then we can rule out any privacy concerns without having to think too much about potential threat models.

threebithacker:
Fixed in #223

Issue #152, opened 2022-12-26
catshadow client_docker_test failure due to retransmission


mixmasala:
```
17:14:47.377 DEBU catshadow: Saving statefile.
17:14:47.385 DEBU catshadow: readInbox command d80c9f2d221ce74897213c56afbc03d1 sent
    client_docker_test.go:764: Message received by b2
    client_docker_test.go:776: 
        	Error Trace:	/go/katzenpost/catshadow/client_docker_test.go:776
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 3
        	Test:       	TestDockerRenameContact
--- FAIL: TestDockerRenameContact (100.41s)
17:14:47.394 DEBU 
```

The test is just counting received messages, and a retransmission due to lost surb-ack causes the number of total messages received to not match the expected number.

leif:
different but related failure, from CI today:
```
2023-01-03T00:47:05.2737920Z 00:47:05.273 INFO provider2_client: OnACK with SURBID [6470d1e30984890b13bedf0ba9c695c3]
2023-01-03T00:47:05.2874683Z 00:47:05.286 DEBU provider2_client: doSend [28d70a0fc9b519e4ce517d333f1da8bc] without SURB
2023-01-03T00:47:05.2875499Z 00:47:05.286 DEBU minclient:de7600004425a919eb7b6f45c47d99c2c97c712d47061dcf87c8d0a8ff02b1ce@provider2: Hop[0] 'provider2' - 410 ms
2023-01-03T00:47:05.2876135Z 00:47:05.286 DEBU minclient:de7600004425a919eb7b6f45c47d99c2c97c712d47061dcf87c8d0a8ff02b1ce@provider2: Hop[1] 'mix1' - 91 ms
2023-01-03T00:47:05.2876721Z 00:47:05.286 DEBU minclient:de7600004425a919eb7b6f45c47d99c2c97c712d47061dcf87c8d0a8ff02b1ce@provider2: Hop[2] 'mix2' - 216 ms
2023-01-03T00:47:05.2877327Z 00:47:05.286 DEBU minclient:de7600004425a919eb7b6f45c47d99c2c97c712d47061dcf87c8d0a8ff02b1ce@provider2: Hop[3] 'mix3' - 156 ms
[...]

PR #223, opened 2023-02-09
remove extra pkiclient instances and unused arguments

mixmasala:
this also removes client's internal pkicache client. Is there any good reason to keep a LRU cache of 8 documents and add extra workers and complexity? I thought not, so this PR removes it, and instead uses a single pki.Client instance to bootstrap and give to minclient.

this also gets rid of some extra delays (we dont need to wait 5 seconds on each pkiclient startup).
cumulatively this seems to make the CI suite run in about half the time. :-)

there were a few places where we were passing DataDir around un-necessarily, so this is now gone as well.


mixmasala:
the last commit should fix the DockerRenameContact test that was failing often :-)

catshadow/client_docker_test.go
```diff
@@ -702,9 +704,22 @@ loop3:
 		}
 	}
 
+	// wait for message to be received by b
+loop3a:
```
threebithacker, line 708: This fixes the symptom (test failing), but not the cause (temporal ordering / concurrency issues in `client.go`).
I suggest we make this change, and then add another RenameContact test that doesn't force the serialization order.
That is, we can't guarantee that the way this change does it is how it plays out in real life.

catshadow/client_docker_test.go
```diff
@@ -785,22 +817,7 @@ loop7:
...
-			} else {
-				if bytes.Equal(last.Plaintext, msg.Plaintext) {
-					t.Logf("%s was retransmitted", last.Plaintext)
-				}
-			}
-		}
-	} else {
-		require.Equal(received, 2)
-	}
+	require.Equal(received, 2)
```
threebithacker, line 820: Wondering if we should assert (for sanity) that:
```golang
require.Equal(1, len(a.conversations))
require.Equal(1, len(b.conversations))
```
It seems like a reasonable requirement, since there is only one conversation going on.

client/session.go
```diff
@@ -128,12 +120,12 @@ func NewSession(
...
 		LinkKey:             s.linkKey,
 		LogBackend:          logBackend,
-		PKIClient:           pkiCacheClient,
+		PKIClient:           pkiClient,
 		OnConnFn:            s.onConnection,
 		OnMessageFn:         s.onMessage,
 		OnACKFn:             s.onACK,
 		OnDocumentFn:        s.onDocument,
-		DialContextFn:       proxyCfg.ToDialContext("authority"),
+		DialContextFn:       proxyCfg.ToDialContext(string(idHash[:])),
```
threebithacker, line 128: This will get passed in plaintext to Tor/socks, maybe we shouldn't leak our internal identifiers to the network.
Since this is only initialized once per session, it seems safer to just generate a random string?
Then we can close https://github.com/katzenpost/katzenpost/issues/56

threebithacker (changes requested):

threebithacker:
Seems related: https://github.com/katzenpost/katzenpost/issues/152

catshadow/client_docker_test.go
mixmasala: Good idea, thanks 923b397542a8323594455f3826ad9019dc6d1791

client/session.go
mixmasala: good catch 153f4b6b37892eaa7036b3329a47798977c80347

catshadow/client_docker_test.go
mixmasala: I'm pretty sure the test failure is because of a bug in the test framework and there isn't a fix to client.go that would stop this problem, and I think the message storage / contact name string should not be used as a key for any of these datastructures... 

catshadow message storage needs to be redesigned; also it serializes the entire message history and writes to disk on every state change.

mixmasala:
I found another pki.Client being instantiated for session. Did we really need to have 3 instances of pki.Client ? I feel like I'm missing something here...

PR #225, opened 2023-02-10
Blinded25519: Implement (un)marshalling for BlindedPrivateKey

threebithacker:


threebithacker:
Blocked on https://github.com/katzenpost/katzenpost/pull/226 since the rules require CI to run

david415 (approved):

core/crypto/eddsa/blinded25519.go
```diff
@@ -41,20 +32,98 @@ package eddsa
...
+		0xf7, 0xa2, 0xde, 0xf9, 0xde, 0x14, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+		0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10}
+	order_sc, _ := new(edwards25519.Scalar).SetUniformBytes(order_64[:])
+	pkA_n , err := new(edwards25519.Point).SetBytes(pk.Bytes())
+	if nil != err {
+		panic("failed to set edwards25519.Point.SetBytes(), incorrect size?")
+	}
+	pkA_n.ScalarMult(order_sc, pkA_n)
+	pk_is_1 := subtle.ConstantTimeCompare(identity_element[:], pk.Bytes())
+	copy(result[:], pkA_n.Bytes())
```
mixmasala, line 67: copy can be omitted I believe,
e.g. 
```
pk_mult_L_is_1 := subtle.ConstantTimeCompare(identity_element[:], pkA_n.Bytes())
```

core/crypto/eddsa/blinded25519.go
```diff
@@ -41,20 +32,98 @@ package eddsa
...
+// Sanity checking of public keys.
+// We do NOT do check for small-order points here or otherwise validate the point.
+// This function is just here to catch accidentally-bad keys, basically.
+// Checks that p != G
+// Checks that p != 1
+// Checks that p != 0
+// Checks that L*p = 1
+func CheckPublicKey(pk *PublicKey) bool {
+	var result [32]byte
+	order_64 := [64]byte{0xed, 0xd3, 0xf5, 0x5c, 0x1a, 0x63, 0x12, 0x58, 0xd6, 0x9c,
```
mixmasala, line 57: maybe put this in scope outside function body and comment as in core/crypto/edwards25519
// order is the order of Curve25519 in little-endian form.
var order = [4]uint64{0x5812631a5cf5d3ed, 0x14def9dea2f79cd6, 0, 0x1000000000000000}

core/crypto/eddsa/blinded25519.go
```diff
@@ -41,20 +32,98 @@ package eddsa
...
+
+// Unmarshal 32 bytes to a private key. Rederives the public key.
+func (k *BlindedPrivateKey) UnmarshalBinary(data []byte) error {
+	if len(data) != 32 {
+		e := errors.New("BlindedPrivatekey.UnmarshalBinary: len(data) != 32")
+		if e == nil { panic("what") }
+		return e
+	}
+	// we could reuse the old k.blinded, but it's easier to throw it away:
+	k.blinded = make([]byte, 64)
```
mixmasala, line 103: this doesn't zero key material before throwing it away though. look at ../core/utils/explicitBzero.go
if the underlying BlindedPrivateKey was allocated as a memguard.LockedBuffer or something it's nicer if it keeps the same backing / slice

mixmasala (approved):
a few small style changes if you please

core/crypto/eddsa/blinded25519.go
mixmasala: (you could also make var order Order (public) from core/crypto/edwards25519 and use that)

threebithacker: I think we should delete `core/crypto/edwards25519.go`. I thought we already did. I'll make a PR for that later.

core/crypto/eddsa/blinded25519.go
threebithacker: `edwards25519` which we use heavily copies internally, as does `crypto/ed25519`, so I think it would be misleading to sprinkle selective deletions throughout the code.
We could use the runtime to clear heap values after free, but I don't think it belongs in this PR.

threebithacker:
The coverage summary is totally flaky, but clicking on the link did show how some of the error paths weren't being tested. I pushed tests for the `MarshalBinary`/`UnmarshalBinary` methods.

PR #235, opened 2023-02-14
Csidh nike wrapper type for the nobs csidh-512

david415:
Add NIKE wrapper type for the nobs pure golang CSIDH-512 implementation that is currently already being used by katzen/doubleratchet.

Change NIKE interface to more closely match the cloudflare/circl KEM interface.

Update all the things that used the NIKE interface... to use the new interface.

Currently the only user of this NIKE interface is core/sphinx
however the double ratchet can use the NIKE interface and not have any memguard like in this branch over here: https://github.com/katzenpost/doubleratchet/pull/16

which is in sync with these latest changes in the CTIDH cgo:
https://github.com/katzenpost/ctidh_cgo/pull/4


core/crypto/nike/csidh/csidh.go
```diff
@@ -0,0 +1,263 @@
...
+	return csidh.PrivateKeySize
+}
+
+func (e *CsidhNike) GenerateKeyPair() (nike.PublicKey, nike.PrivateKey, error) {
+	privateKey := new(csidh.PrivateKey)
+	err := csidh.GeneratePrivateKey(privateKey, rand.Reader)
+	if err != nil {
+		return nil, nil, err
+	}
+	publicKey := new(csidh.PublicKey)
```
threebithacker, line 55: looks like this could use `DerivePublicKey()` to prevent code duplication?

core/crypto/nike/nike.go
```diff
@@ -78,6 +89,21 @@ type Nike interface {
 	// See also PublicKey's Blind method.
 	Blind(groupMember []byte, blindingFactor []byte) (blindedGroupMember []byte)
```
threebithacker, line 90: Why is this `[]byte` and not just `groupMember *PublicKey, blindingFactor *PrivateKey` ? Then we can centralize the validation, and it also becomes less unclear how to come up with valid blinding factors.
Looks like it should also return a `*PublicKey` here?

core/crypto/nike/nike.go
```diff
@@ -46,18 +54,21 @@ type PublicKey interface {
...
 	// PublicKeySize returns the size in bytes of the public key.
 	PublicKeySize() int
 
 	// PrivateKeySize returns the size in bytes of the private key.
 	PrivateKeySize() int
 
-	// NewKeypair returns a newly generated key pair.
-	NewKeypair() (PrivateKey, PublicKey)
+	// GenerateKeyPair creates a new key pair.
+	GenerateKeyPair() (PublicKey, PrivateKey, error)
```
threebithacker, line 71: It would be nice if this took a `io.Reader` argument so it can be called/tested deterministically instead of using `rand.Reader` under the hood.
Same goes for `Blind()` which also calls `rand.Reader`

core/crypto/nike/nike.go
```diff
@@ -46,18 +54,21 @@ type PublicKey interface {
 	Blind(blindingFactor []byte) error
```
threebithacker, line 54: This here should also just take a `*PrivateKey`

threebithacker (changes requested):

core/crypto/nike/csidh/csidh.go
david415: yeup DerivePublicKey fits here, thanks

core/crypto/nike/nike.go
david415: We've actually got two Blind ops. The math is the same... both are for the public key blinding but one is called from the Scheme reference and the other is a method of PublicKey. The PublicKey method mutates the public key. Other Scheme method simply returns the blinded value. We have it like this and their arguments take the types they currently take to make it convenient to from this bit of code here:

https://github.com/katzenpost/katzenpost/blob/main/core/sphinx/sphinx.go#L332-L345

david415: But @threebithacker I'm open to changing the function signatures but that shouldn't block this pull request. We can do that in another pull request. I mean... @leif already complained that this pull request is 17 files. Wouldn't want to make it even bigger.

core/crypto/nike/nike.go
david415: i completely agree. however there might be some value in keeping this method as well since it conforms to the cloudflare circl's KEM Scheme interface https://github.com/cloudflare/circl/blob/main/kem/kem.go#L38

What I'm saying is that it was 100 % an intentional act to make our NIKE interfaces very similar to the KEM interfaces.

core/crypto/nike/nike.go
david415: When using the Blind method to implement Sphinx, we have a raw byte slice, not a private key!
Here is where we derive the blinding factors:

https://github.com/katzenpost/katzenpost/blob/main/core/sphinx/internal/crypto/crypto.go#L181-L182

This code has been like this for years.
You can't block this PR because of this.
If you want me to change it... I can do that in another pull request.

core/crypto/nike/nike.go
david415: hopefully this is acceptable to you:
```
	// GenerateKeyPair creates a new key pair.
	GenerateKeyPair() (PublicKey, PrivateKey, error)

	// GenerateKeyPairFromEntropy creates a new key pair from the given entropy source.
	GenerateKeyPairFromEntropy(rng io.Reader) (PublicKey, PrivateKey, error)

```

david415: OKay I've added the other constructor and I made it work for CTIDH C bindings

PR #239, opened 2023-02-18
Add status to catshadow

mixmasala:
Adds Status() command to indicate whether client is connected/connecting or offline.


catshadow/client.go
```diff
@@ -1378,30 +1385,29 @@ func (c *Client) Online() error {
 // goOnline is called by worker routine when a goOnline is received. currently only a single session is supported.
 func (c *Client) goOnline() error {
 	c.connMutex.RLock()
-	if c.online || c.connecting || c.session != nil {
+	if c.connected == StateOnline || c.connected == StateConnecting {
 		c.connMutex.RUnlock()
 		return errors.New("Already Connected")
 	}
 	c.connMutex.RUnlock()
```
threebithacker, line 1393: There's a race here where after `RUnlock()` and before `c.connMutex.Lock()` the state could change, meaning two `goOnline()` can be running at the same time, launching two parallel `c.client.NewTOFUSession()`s.
Since `sync` doesn't seem to have an *upgradeable lock* (???) we could roll our own by having an exclusive mutex guard lock acquisition for the `sync.RWLock` and chuck that in our `utils` library.
In this specific case though, I think it seems better to just take the write Lock up front:
```go
c.connMutex.Lock() // could also be a TryLock where we return early if it fails?
switch c.connected {
    case StateOnline: {
       c.connMutex.Unlock()
       return errors.New("Already Online")
    }
    case StateConnecting: {
       c.connMutex.Unlock()
       return errors.New("Already Connecting")
    }
}
c.connected = StateConnecting
c.connMutex.Unlock()
```
Alternatively I guess we could explore using atomics for state transitions?

catshadow/client.go
```diff
@@ -1441,16 +1447,16 @@ func (c *Client) getSpoolWriteDescriptor() *memspoolclient.SpoolWriteDescriptor
 func (c *Client) goOffline() error {
 	c.connMutex.Lock()
 	defer c.connMutex.Unlock()
-	if c.connecting {
+	if c.connected == StateConnecting {
 		return errors.New("Offline() does not cancel Online()")
 	}
 
-	if !c.online || c.session == nil {
```
threebithacker, line 1448: what happened to the `c.session` check, is it an error if we are `StateConnecting` and have a `nil` session? or `StateOffline` but a non-`nil` `c.session`?

threebithacker (changes requested):

catshadow/client.go
mixmasala: see d44d376bf51c93c49f51eed1bab622df7599b269

mixmasala: (thanks)

catshadow/client.go
mixmasala: I think it should never be in StateConnecting with a non-nil session

mixmasala: the check is removed since doOnline does the same check, and there is only one place where session is set or nil'd while the lock is held

threebithacker: Hmm I can't find `doOnline` anywhere, which file is that in?

threebithacker: Sorry it's not: because `if c.connected == StateConnecting { return errors `

mixmasala: Is this resolved or can you propose a change to be made?

catshadow/client.go
threebithacker: https://github.com/katzenpost/katzen/pull/29#issuecomment-1595825465
new code takes a read-only lock `RLock()` then tries to relinquish a read-write lock (`Unlock()`), this results in a crash

leif: > [katzenpost/katzen#29 (comment)](https://github.com/katzenpost/katzen/pull/29#issuecomment-1595825465) new code takes a read-only lock `RLock()` then tries to relinquish a read-write lock (`Unlock()`), this results in a crash

fixed in c3a210d0ae74ec152cf6807b45aa50ee56d74b15

PR #245, opened 2023-02-26
Sphinx api upgrade

david415:
Upgrade the Sphinx API such that the geometry is the only information you need to create a Sphinx type. Therefore the Sphinx Geometry functions as a factory for the Sphinx type.


leif:
(reopening so that when #246 lands this shows up as merged instead of appearing abandoned)

Issue #182, opened 2023-01-02
panic: failed to create session: document has signature by unknown verifier


leif:
This happens if a client comments out one of the authorities in their config (or if a new authority was added and the client hadn't upgraded). Probably we should remove this panic and accept unknown verifiers (if a majority of the known ones have signed it) so that adding new authorities won't require clients to upgrade.

leif:
This issue also means that adding a new authority without network downtime requires a majority of the authorities and a viable number of providers and mixes to all upgrade within the same epoch.

The relevant place to fix this is here:
https://github.com/katzenpost/katzenpost/blob/2ad5ffff92d68f8f4e6d86dc02bd9e6684b5e70d/core/pki/document.go#L418

Glancing at it now, unless I'm misunderstanding something it appears that it also will give up verifying if there is a single bad signature? That should probably also be fixed, even though there are still other ways that a single bad authority can prevent the good ones from reaching any consensus we don't need to be *this* fragile. (I think there are several other similar fragilities in the same file...)

PR #248, opened 2023-03-05
Slightly improve PKI client cryptographic processing of the doc

david415:


leif (approved):
LGTM. Everywhere `VerifyAndParseDocument` was called, the document has already been verified. This also fixes #182

Issue #1, opened 2021-09-06
monorepo migration


leif:
This is the first issue in the new Katzenpost monorepo. Please link to any related issues here.

We should close this issue when these tasks are done:

- [ ] Update website from monorepo instead of old website repo
- [ ] Write blog post announcing the switch to a monorepo
- [ ] Archive the repos that have been merged into this repo
- [ ] Transfer open issues from those repos to this one using github's "Transfer issue" feature.
- [ ] Write and document a script for developers to add old repos as a remote on their monorepo, configured to not pull tags and to be able to pull PRs so that they can be merged. (Merges should from multiple repos to this one will work automatically using git locally, but not using github's interface...)

leif:
Because this monorepo contains the full histories of the old `master` branches from many different katzenpost repos, it is possible to pull from them and merge the PRs without many (or sometimes any) conflicts as long as they were in a mergeable state prior to the monorepo migration and the relevant files have not changed here since then.

To turn a PR like https://github.com/katzenpost/server/pull/151 from another repo into a branch named `server_pr151` in this monorepo, in my local checkout of https://github.com/katzenpost/katzenpost I ran this command:
```
git fetch --no-tags https://github.com/katzenpost/server/ pull/151/head:server_pr151
```

It is important to use `--no-tags` when fetching from other repos, as we do not want their tags here.

I think ran `git merge main` to see if it would merge cleanly. In the case of this PR, there were three conflicts, due to commits in the PR changing the order of some imports near other internal imports which the monorepo migration had to modify. I manually resolved those, and opened https://github.com/katzenpost/katzenpost/pull/5 and closed https://github.com/katzenpost/server/pull/151.

david415:
All our design docs have URLs in the references to the other design docs. All those URLs are now incorrect. Furthermore at the end of each doc there's a bibtex quote that people can use to cite our documents. Those URLs are now incorrect.

david415:
Some of the URLs didn't get "migrated":

Look at the bibtex quote at the bottom of this one: https://github.com/katzenpost/katzenpost/blob/main/docs/specs/mixnet.rst

There might be more like this. I didn't check.

david415:
Okay i just checked:
human@computer ~/gopath/src/github.com/katzenpost/katzenpost (main) $ grep -r https://github.com/Katzenpost/docs | wc -l
26

26 of the URLs didn't get migrated

david415:
[...]

PR #250, opened 2023-03-06
Use NIKE interface in doubleratchet

leif:
from https://github.com/katzenpost/doubleratchet/pull/16


leif:
note: we want to test with katzen to ensure there is an upgrade path before merging this.

we'll have some breaking changes soon enough, but hopefully this one doesn't need to be.

leif (commented):
tagging https://github.com/katzenpost/katzenpost/issues/1 to note that https://github.com/katzenpost/doubleratchet/pull/16 was migrated to https://github.com/katzenpost/katzenpost/pull/250 in this commit

ioerror:
Is this still relevant and should it be updated to include `highctidh` that has just landed on master?

PR #253, opened 2023-04-24
cryptography security fix, code quality improvement

david415:
Please note that I also had to change our fork of the nyquist golang noise library by addind the `FromKEM` function to allow us to pass in whatever KEM we want, here:

https://github.com/katzenpost/nyquist/pull/2

Use an IND-CCA2 secure KEM instead of a IND-CPA KEM with noise. Therefore without more closely inspecting the Noise protocol, it's difficult to say if we achieve semantic security with an IND-CPA KEM used with Noise pqXX. It seems unlikely to achieve IND-CCA2 given that Noise uses a separate hash objects for ciphertext vs public keys. Whereas Cloudflare had the intention of using IND-CPA KEMs with TLS, which would be fine because TLS uses one hash object and achieves IND-CCA2 security.

The fix is two-fold:
1. NIKE to KEM adapter with IND-CCA2 security
The adapter itself is simple. It's just a DH operation masquerading as a KEM. However here we make it IND-CCA2 by using a hash function to mix in various values, ciphertext/public keys.

2. security preserving KEM combiner
We base our KEM combiner off the one in Cloudflare's circl library and make the small modification of adding the use of a hash function in the Encapsulate and Decapsulation methods to achieve IND-CCA2. I did manage to get a review from a friendly cryptographer https://github.com/cloudflare/circl/issues/432 who suggested the `H(ss1 || ss2 || ct1 || ct2)` construction for the "split PRF".

The plan is therefore to make a hybrid KEM with Kyber768 and use our new security preserving KEM combiner to combine it with the NIKE to KEM adapter with X25519, which is just a KEM as well. This results in our IND-CCA2 hybrid post quantum KEM, suitable for use with the Noise cryptographic protocol framework, defined like this:

```
		hybrid.New(
			"Kyber768-X25519",
			adapter.FromNIKE(ecdh.NewEcdhNike(rand.Reader)),
			kyber768.Scheme(),
		),
```

And in addition to all that I had to fix a bit of technical debt in our "core/wire" where the Noise message sizes were hardcoded, now they are parameterized on the KEM scheme so that it always calculates the correct message sizes based on the KEM's ciphertext size and public key size.

I've also added more specific types for CTIDH such that each key size has it's own type and therefore we have more specific types for the hybrid NIKE that you see here:

```
var CTIDH1024X25519 nike.Scheme = &scheme{
	name:   "CTIDH1024-X25519",
	first:  ctidh.CTIDH1024Scheme,
	second: ecdh.NewEcdhNike(rand.Reader),
}
```


docs/specs/kemsphinx.rst
```diff
@@ -30,7 +32,69 @@ has two functions:
...
+```
+func Encap(their_pubkey publickey) ([]byte, []byte) {
+        my_privkey, my_pubKey = GEN_KEYPAIR(RNG)
+        ss = DH(my_privkey, their_pubkey)
+        ss2 = H(ss || my_pubkey || their_pubkey)
+	return my_pubkey, ss2
+}
+
+func Decap(my_privkey, ciphertext) []byte {
+        s = DH(my_privkey, ciphertext)
```
threebithacker, line 66: 1. The `DH` used here is unclear. In `Encap` it' used like old-school DH with public+private keys, in `Decap` it takes `ciphertext` ? What is `ciphertext` here? Is this `KEM-DECRYPT` rather than DH or should `ciphertext` have read `their_pubkey` ?
2. `Encap` and `Decap` as outlined here produce different secrets:
    ```
    Encap_ss2 = H(ss || a_pubkey || b_pubkey)
    Decap_ss2 = H(ss || b_pubkey || ciphertext) # ciphertext=a_pubkey given 1)
    ```
    the order of the pubkeys should be flipped in one of them so it's consistent.

docs/specs/kemsphinx.rst
```diff
@@ -30,7 +32,69 @@ has two functions:
...
+	shared_key = H(ss || my_pubkey || ciphertext)
+	return shared_eky
```
threebithacker, line 68:
```suggestion
	return shared_key
```

docs/specs/kemsphinx.rst
```diff
@@ -30,7 +32,69 @@ has two functions:
...
+two KEM shared secrets together we use a hash function to also mix
+in the values of both KEM ciphertexts:
+
+```
+// H is a hash function
+
+func splitPRF(ss1, ss2, cct1, cct2 []byte) []byte {
+        return H(ss1 || ss2 || cct1 || cct2)
+}
+```
```
threebithacker, line 94: This seems good, but `cct1` and `cct2` should be defined explicitly in `Encap`/`Decap`: it should return a `cct` value, the concatenation of the public keys in a defined order; that way the combiner doesn't have to violate the layers.

threebithacker: For this to be a PRF I believe it would need to take an additional parameter to parameterize the session key over?
Like we'd want to use it something like
 `recv_key, send_key := splitPRF(blabla, "server"), splitPRF(blabla, "client")`

core/wire/noise_test.go
```diff
@@ -38,18 +43,19 @@ func TestNyquistPqNoiseParams2(t *testing.T) {
...
-		Hash:    hash.BLAKE2s,
+		KEM: kem.FromKEM(
+			kemhybrid.New(
+				"Kyber768-X25519",
+				adapter.FromNIKE(ecdh.NewEcdhNike(rand.Reader)),
+				kyber768.Scheme(),
+			),
+		),
+		Cipher: cipher.ChaChaPoly,
+		Hash:   hash.BLAKE2s,
```
threebithacker, line 56: Do we have blake2b? I think it'd be interesting to see a benchmark between the two; as I understand it blake2s is the microcontroller-optimized algorithm, while we are probably targeting 64-bit machines here?
If we're changing all the other things we might as well deal with this now?

core/crypto/nike/hybrid/not_ctidh.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/katzenpost/katzenpost/core/crypto/rand"
 )
 
-var CTIDHX25519 nike.Scheme = &scheme{
+var CTIDH1024X25519 nike.Scheme = &scheme{
```
threebithacker, line 12: What is the reason for calling this `CTIDH1024X255129` and `CTIDH-X25519` if it doesn't use CTIDH at all?
It would be good with a comment in here explaining what the purpose of this file is.

core/crypto/nike/hybrid/hybrid.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/katzenpost/katzenpost/core/crypto/rand"
 )
 
-var NOBS_CSIDHX25519 nike.Scheme = &scheme{
+var NOBS_CSIDH512X25519 nike.Scheme = &scheme{
 	name:   "NOBS_CSIDH-X25519",
```
threebithacker, line 14: Should this name be `NOBS_CSIDH512-X25519`? I don't understand the relation between the variable name and the scheme `name` parameter in these. In `ctidh.go` you updated all three to have the same name

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,249 @@
...
+			scheme:    a,
+		}, &PrivateKey{
+			privateKey: privkey,
+			scheme:     a,
+		}, nil
+}
+
+// Encapsulate generates a shared key ss for the public key and
+// encapsulates it into a ciphertext ct.
+func (a *Scheme) Encapsulate(pk kem.PublicKey) (ct, ss []byte, err error) {
```
threebithacker, line 103: this is `ENCAP` in the `kemsphinx.rst` document if I'm not mistaken - we should have a note about that.
Additionally we should consider renaming it to `ENCAPSULATE` in the spec so the names match.

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,249 @@
...
+	seed := make([]byte, a.EncapsulationSeedSize())
+	_, err = rand.Reader.Read(seed)
+	if err != nil {
+		return
+	}
+	return a.EncapsulateDeterministically(pk, seed)
+}
+
+func h(publicKeySize int, ss []byte, pubkey1 []byte, pubkey2 []byte) []byte {
```
threebithacker, line 112: let's call this something greppable like `hashSharedSecretWithPublicKeys` or whatever. It's not a utility function we use very often, and this file has a lot of local `h` pointers to hash function class instances.

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// Name of the scheme
+func (a *Scheme) Name() string {
+	return a.nike.Name()
+}
+
+// GenerateKeyPair creates a new key pair.
+func (a *Scheme) GenerateKeyPair() (kem.PublicKey, kem.PrivateKey, error) {
+	pubkey, privkey, err := a.nike.GenerateKeyPair()
```
threebithacker, line 88: Do we want deterministic functions for these like we have `GenerateKeyPairFromEntropy` ? Seems like it could be a useful building block for other interesting constructions, but especially for testing.

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,249 @@
...
+	pk2, sk2 := a.DeriveKeyPair(seed)
+	ss := a.nike.DeriveSecret(sk2.(*PrivateKey).privateKey, pub.publicKey)
+	ss2 := h(a.nike.PublicKeySize(), ss, pub.publicKey.Bytes(), pk2.(*PublicKey).publicKey.Bytes())
+	ct, _ := pk2.MarshalBinary()
+	return ct, ss2, nil
+}
+
+// Size of seed used in EncapsulateDeterministically().
+func (a *Scheme) EncapsulationSeedSize() int {
+	return a.nike.PublicKeySize()
```
threebithacker, line 248: Shouldn't this be `SeedSize` ?

It looks like it calls this function:
```go
// DeriveKeyPair deterministicallly derives a pair of keys from a seed.
// Panics if the length of seed is not equal to the value returned by
// SeedSize.
func (a *Scheme) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
	if len(seed) != a.SeedSize() {
```

I guess for testing whether we got this right we could have a `HeteregenousLengthTestScheme` with distinct lengths for each so we can identify places where we screw this up before slipups make it into the rest of our codebase?

core/crypto/kem/adapter/kem_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	t.Logf("hello my name is %s", s.Name())
+
+	pubkey1, privkey1, err := s.GenerateKeyPair()
+	require.NoError(t, err)
+
+	ct, ss, err := s.Encapsulate(pubkey1)
+	require.NoError(t, err)
+
+	ss2, err := s.Decapsulate(privkey1, ct)
```
threebithacker, line 24: using `ss` and `ss2` here is confusing when the functions internally have `ss` and `ss2`. can we call them `ss_a` and `ss_b` here or something like that?

threebithacker (changes requested):

core/crypto/kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,341 @@
...
+
+	return &PublicKey{sch, pk1, pk2}, &PrivateKey{sch, sk1, sk2}
+}
+
+func splitPRF(ss1, ss2, ct1, ct2 []byte) []byte {
+	// implement split PRF KEM combiner as:
+	// H(ss1 || ss2 || ct1 || ct2)
+	// in order to retain IND-CCA2 security
+	// as described here:
+	// https://link.springer.com/chapter/10.1007/978-3-319-76578-5_7
```
threebithacker, line 209: At a first glance the extra hashing in of `ct1, ct2` here seems pointless. Does the paper suggest that it's stronger this way? Seems to me that putting attacker-controlled inputs directly into the combiner here would be riskier than the hash-of-hash product, but I'd be interesting to read what they say.

The paper referenced has a lot of different constructions; we should write explicitly which one we are trying to implement, with a page reference and a unique identifier.
Otherwise the reviewer has to read a 30 page cryptography paper and then try to match an implementation, which may or may not actually match something in the paper, to any of the constructions there. I'm not going to do that today. :-)

docs/specs/kemsphinx.rst
```diff
@@ -19,7 +19,9 @@ route length hiding.
 
 We'll express our KEM Sphinx header in pseudo code. The Sphinx body
 will be exactly the same as [SPHINXSPEC]_. Our basic KEM API
-has two functions:
+has three functions:
+
+* privkey, pubkey = GEN_KEYPAIR(RNG)
```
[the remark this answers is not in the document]
david415: fixed

docs/specs/kemsphinx.rst
david415: here ciphertext is just another public key but i will rename it so that it's more clear

david415: fixed

docs/specs/kemsphinx.rst
david415: fixed

core/wire/noise_test.go
david415: Yeah I think we mostly care about 64-bit machines. blake2b is probably a fine choice.

core/crypto/nike/hybrid/not_ctidh.go
david415: that file only makes sense if you read the build tag at the top and look at the other files in the same dir and look at their build tags as well.

core/crypto/nike/hybrid/hybrid.go
david415: the name string allows you to get the schema with a query with that string name from the "kem/schemes" module

core/crypto/kem/adapter/kem.go
david415: fixed. set it to 32

core/crypto/kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,341 @@
...
+	if len(seed) != sch.EncapsulationSeedSize() {
+		return nil, nil, kem.ErrSeedSize
+	}
+
+	h := sha3.NewShake256()
+	_, _ = h.Write(seed)
+	first := make([]byte, sch.first.EncapsulationSeedSize())
+	second := make([]byte, sch.second.EncapsulationSeedSize())
+	_, _ = h.Read(first)
+	_, _ = h.Read(second)
```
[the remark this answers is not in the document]
david415: You'll notice there's a LICENSE file in this directory. That's because I didn't write it... but I'm making a few small modifications to it so that the types are exported and so that it's IND-CCA2 instead of IND-CPA.

Please see cloudflare's original hybrid kem combiner, here: https://github.com/cloudflare/circl/blob/main/kem/hybrid/hybrid.go#L262-L290

core/crypto/kem/schemes/kem_test.go
```diff
@@ -0,0 +1,27 @@
...
+	t.Logf("encapsulation seed size %d", s.EncapsulationSeedSize())
+
+	pubkey1, privkey1, err := s.GenerateKeyPair()
+	require.NoError(t, err)
+	ct, ss, err := s.Encapsulate(pubkey1)
+	require.NoError(t, err)
+	ss2, err := s.Decapsulate(privkey1, ct)
+	require.NoError(t, err)
+	require.Equal(t, ss, ss2)
+	t.Logf("our shared key is %x", ss)
```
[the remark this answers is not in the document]
david415: fixed

core/crypto/nike/hybrid/not_ctidh.go
david415: I will remove this since it is not needed... but if you would please next time look at the build tags at the top of the source file you'll see that what's going on here is that we have a hybrid type that changes depending on if ctidh build tag is used or not. this lets us reference this type in the nikes/schemes and kem/schemes modules where we have a lookup by name helper function.

core/crypto/kem/hybrid/hybrid.go
david415: Let's talk about this offline, since a cryptographer already told me to use H(ss1 || ss2 || ct1 || ct2)

docs/specs/kemsphinx.rst
david415: no... this is perfectly correct as is. let's talk elsewhere or have a phone call or something.

core/wire/noise_test.go
david415: lol okay please limit the scope of this code review to the code that i actually wrote in the pull request.

david415:
ready for more review

david415:
please also review https://github.com/katzenpost/ctidh_cgo/pull/5

core/crypto/kem/adapter/kem.go
threebithacker: we now have `DeriveKeyPair` in this file

core/crypto/kem/hybrid/hybrid.go
threebithacker: We're not hashing `H(ss1 || ss2 || ct1 || ct2)` here. The `ss1` and `ss2` parameters to this function are (more or less) `H(ss_x, ct_x)`, thus `ct_x` (the public keys) are hashed in twice.
The linked-to paper is a very long paper and does not appear to have a quote a cryptographer recommending this nested scheme. It's still unclear which scheme from the paper this function was intended to implement.

threebithacker (changes requested):

core/crypto/kem/adapter/kem_test.go
david415: I prefer snake case too... but golang is a camel case language. So we'll have to do ssA and ssB

core/crypto/kem/hybrid/hybrid.go
david415: The current construction as I've written it is perfectly correct but I can see how it's maybe maybe slightly confusing but let me explain. We do NOT care at all what each individual KEM does when we are working on the code inside the KEM combiner... except that we start with the assumption that we are combining two KEMs which both have IND-CCA2 security. The idea here is that if one of the KEMs has an attack whereby it's security is reduced to IND-CPA then in that case this KEM combiner will still retain the IND-CCA2 security. Happy to chat with you about this outside of this pull request.

core/crypto/kem/hybrid/hybrid.go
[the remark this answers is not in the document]
david415: Okay. Fair enough. I'll fix this.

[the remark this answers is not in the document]
david415: fixed forever!

core/crypto/kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,334 @@
...
+		return nil, err
+	}
+	second, err := sk.second.MarshalBinary()
+	if err != nil {
+		return nil, err
+	}
+	return append(first, second...), nil
+}
+
+func (sk *PrivateKey) Equal(other kem.PrivateKey) bool {
```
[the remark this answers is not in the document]
david415: it looks like they are handling pointer nil semantics of equality here. i suppose they had some uses in mind like:

```
var myKey PrivateKey = nil

// marklar marklar...

if myKey == nil {
        // marklar
}

// marklar 

if myKey.Equal(otherKey) {
        // marklar
}
```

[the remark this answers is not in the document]
david415: lol maybe they added that stuff in to prevent a bunch of panics... because golang runtime will panic if you try to compare a byte slice if it is set to nil.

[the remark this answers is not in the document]
david415: so... yeah we can remove that crufty code... but there might be a hidden state machine somewhere that later crashes with a panic inside the Equal method... with a nil pointer dereference error traceback ;-p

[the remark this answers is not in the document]
david415: but in defense of your suggestion that we may not need it... we don't have such checks anywhere else in our code and we've been fine.

threebithacker (changes requested):

core/wire/kem.go
```diff
@@ -27,19 +27,32 @@ import (
...
+	"github.com/katzenpost/katzenpost/core/crypto/nike/ecdh"
 	cpem "github.com/katzenpost/katzenpost/core/crypto/pem"
+	"github.com/katzenpost/katzenpost/core/crypto/rand"
 )
 
 // PublicKeyHashSize indicates the hash size returned
 // from the PublicKey's Sum256 method.
 const PublicKeyHashSize = 32
 
+var KEMScheme = hybrid.New(
```
threebithacker, line 46: Is this `var KEMScheme` exported here - is it used elsewhere? Looks like `DefaultScheme` ought to be enough?

threebithacker (changes requested):

core/crypto/kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,329 @@
...
+
+	return &PublicKey{sch, pk1, pk2}, &PrivateKey{sch, sk1, sk2}, nil
+}
+
+func (sch *Scheme) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
+	if len(seed) != sch.SeedSize() {
+		panic(kem.ErrSeedSize)
+	}
+
+	h, err := blake2b.NewXOF(uint32(sch.first.SeedSize()+sch.second.SeedSize()), nil)
```
threebithacker, line 170: I think it would be nice to just require `len(seed) == sch.first.SeedSize()+sch.second.SeedSize()` and then passing slices on to `sch.XXX.DeriveKeyPair(slice_XXXX)` here, instead of the force `blake2b.NewXOF`  step?

docs/specs/kemsphinx.rst
```diff
@@ -30,7 +32,69 @@ has two functions:
...
+IND-CCA2 security if one of the composing KEMs does not have IND-CCA2
+security. Likewise the paper points out that when using a security
+preserving KEM combiner, if only one of the composing KEMs has
+IND-CCA2 security then the resulting hybrid KEM will have IND-CCA2
+security.
+
+Our KEM combiner uses the split PRF design from the paper when combining
+two KEM shared secrets together we use a hash function to also mix
+in the values of both KEM ciphertexts:
+
```
threebithacker, line 87: - I think here we should mention what `cct1`, `cct2`, `ss1` and `ss2` are, ie they're the values from `ENCAPSULATE()` of the underlying KEMs `1` and `2`, and `cct{1,2}` are the ephemeral envelope public keys ?

docs/specs/kemsphinx.rst
```diff
@@ -30,7 +32,69 @@ has two functions:
...
+We easily achieve our IND-CCA2 security by means of hashing
+together the DH shared secret along with both of the public keys:
+
+```
+func ENCAPSULATE(their_pub_key publickey) ([]byte, []byte) {
+        my_priv_key, my_pub_Key = GEN_KEYPAIR(RNG)
+        ss = DH(my_priv_key, their_pub_key)
+        ss2 = H(ss || my_pubkey || their_pubkey)
+	return my_pub_key, ss2
+}
```
threebithacker, line 63: In our implementation of `core/crypto/kem/adapter/kem.go:EncapsulateDeterministically`
```go
	ss := a.nike.DeriveSecret(sk2.(*PrivateKey).privateKey, pub.publicKey)
        // ss = DH(my_priv_key=sk2 || their_pub_key=pub)
	ss2 := hashSharedSecretWithPublicKeys(a.nike.PublicKeySize(), ss, pub.publicKey.Bytes(), pk2.(*PublicKey).publicKey.Bytes())
        // ss2 = H(ss || their_pub_key=pub || my_pubkey=pk2) <--- the order of their_pub, my_pub is switched here
```
and `core/crypto/kem/adapter/kem.go:Decapsulate`:
```go
        ss := a.nike.DeriveSecret(sk.(*PrivateKey).privateKey, pk.(*PublicKey).publicKey)
	ss2 := hashSharedSecretWithPublicKeys(a.nike.PublicKeySize(), ss, sk.Public().(*PublicKey).publicKey.Bytes(), pk.(*PublicKey).publicKey.Bytes())
       // H(ss || my_pubkey || their_pubkey=pk)
```
The implementation results in matching `ss2 == shared_key`, the specification does not:
```
bob_pub , bob_ss2 = bob.ENCAPSULATE(alice_pub)
alice_ss2 = alice.DECAPSULATE(bob_pub)
// bob_ss2 = H(ss || bob_pub || alice_pub)
// alice_ss2 = H(ss || alice_pub || bob_pub)
```

threebithacker (changes requested):

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,264 @@
...
+func (a *Scheme) Encapsulate(pk kem.PublicKey) (ct, ss []byte, err error) {
+	seed := make([]byte, a.EncapsulationSeedSize())
+	_, err = rand.Reader.Read(seed)
+	if err != nil {
+		return
+	}
+	return a.EncapsulateDeterministically(pk, seed)
+}
+
+func hashSharedSecretWithPublicKeys(publicKeySize int, ss []byte, pubkey1 []byte, pubkey2 []byte) []byte {
```
threebithacker, line 118: I think we should use the naming scheme from the spec `their_pubkey`, `own_pubkey` etc instead of pk1,pk2 here.

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,264 @@
...
+
+// Returns the shared key encapsulated in ciphertext ct for the
+// private key sk.
+// Implements DECAPSULATE as described in NIKE to KEM adapter,
+// see docs/specs/kemsphinx.rst
+func (a *Scheme) Decapsulate(sk kem.PrivateKey, ct []byte) ([]byte, error) {
+	if len(ct) != a.CiphertextSize() {
+		return nil, kem.ErrCiphertextSize
+	}
+	pk, err := a.UnmarshalBinaryPublicKey(ct)
```
threebithacker, line 154: The naming scheme where `sk` belongs to `a` and `pk` belongs to the other party is confusing.

threebithacker (changes requested):

core/crypto/kem/hybrid/hybrid.go
threebithacker: 1. It's "perfectly correct" in the sense that it's not harmful iff the hash function has no issues (length extension attacks, etc).
If it has, `H(H(ss1 || ct1) || H(ss2 || ct2))` would be better since bogus inputs to ct1/ct2 would not be able to directly influence the hashing state of the combined value.
In the single-KEM instance, `H(ct || ss)` is probably more robust than `H(ss || ct)`.

2. It's still unclear to me on what page of the referenced paper this scheme is discussed.

threebithacker (changes requested):

core/crypto/kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,329 @@
...
+		panic(err)
+	}
+	_, err = h.Write(ct2)
+	if err != nil {
+		panic(err)
+	}
+	return h.Sum(nil)
+}
+
+func (sch *Scheme) Encapsulate(pk kem.PublicKey) (ct, ss []byte, err error) {
```
threebithacker, line 217: This implementation can just be a thin wrapper around `EncapsulateDeterministically` like in other implementations, then we don't risk the implementations diverging.

core/crypto/kem/hybrid/hybrid.go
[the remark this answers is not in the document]
threebithacker: `blake2b.XOF` is definitely better than `sha3` here, but I still think we could just ask for a larger random chunk (as described in other comment in this review), then the caller can deal with all the direct inputs.

core/crypto/nike/hybrid/no_nobs_csidh.go
```diff
@@ -9,8 +8,8 @@ import (
...
 )
 
-var CTIDHX25519 nike.Scheme = &scheme{
-	name:   "CTIDH-X25519",
-	first:  nil,
-	second: ecdh.NewEcdhNike(rand.Reader),
+var NOBS_CSIDHX25519 nike.Scheme = &scheme{
+	name:   "NOBS_CSIDH-X25519",
+	first:  ecdh.NewEcdhNike(rand.Reader),
+	second: nil,
```
threebithacker, line 14: Since this isn't using CSIDH on `ppc64le`, what's the point of defining it?
Ie we should have static test vectors for this, and those would then fail on `ppc6le`.
Seems better to have it unavailable on `ppc64le` altogether?

core/sphinx/sphinx_benchmark_test.go
```diff
@@ -28,8 +28,16 @@ import (
...
+	if mynike == nil {
+		panic("nike cannot be nil")
+	}
+
 	g := geo.GeometryFromUserForwardPayloadLength(mynike, len(testPayload), false, 5)
-	sphinx := NewSphinx(g)
+	sphinx := NewNIKESphinx(mynike, g)
+
+	if sphinx.nike == nil {
+		panic("nike cannot be nil")
```
threebithacker, line 39:
```suggestion
		panic("sphinx.nike cannot be nil")
```
(to disambiguate from above crash message)

core/sphinx/sphinx_ctidh_test.go
```diff
@@ -31,26 +36,43 @@ func TestHybridCtidhForwardSphinx(t *testing.T) {
...
 }
 
+func TestSphinxConstruction(t *testing.T) {
+	var mynike nike.Scheme
+	mynike = ecdhnike.NewEcdhNike(rand.Reader)
+	g := geo.GeometryFromUserForwardPayloadLength(mynike, 12345, false, 5)
+	t.Logf("NIKEName %s", g.NIKEName)
+	sphinx := NewSphinx(g)
+	require.NotNil(t, sphinx.nike)
+
```
threebithacker, line 57: Could return early here on those platforms?

core/sphinx/sphinx_ctidh_test.go
```diff
@@ -59,20 +81,18 @@ func TestCtidhSURB(t *testing.T) {
...
 
 func TestCTIDHSphinxGeometry(t *testing.T) {
 	t.Parallel()
 	withSURB := false
-	g := geo.GeometryFromUserForwardPayloadLength(ctidhnike.CTIDHScheme, 512, withSURB, 5)
+	g := geo.GeometryFromUserForwardPayloadLength(ctidh.CTIDH1024Scheme, 512, withSURB, 5)
 	t.Logf("NIKE Sphinx CTIDH 5 hops: HeaderLength = %d", g.HeaderLength)
-	g = geo.GeometryFromUserForwardPayloadLength(ctidhnike.CTIDHScheme, 512, withSURB, 10)
+	g = geo.GeometryFromUserForwardPayloadLength(ctidh.CTIDH1024Scheme, 512, withSURB, 10)
 	t.Logf("NIKE Sphinx CTIDH 5 hops: HeaderLength = %d", g.HeaderLength)
```
threebithacker, line 97:
```suggestion
	t.Logf("NIKE Sphinx CTIDH 10 hops: HeaderLength = %d", g.HeaderLength)
```

threebithacker (changes requested):

core/crypto/kem/hybrid/hybrid.go
david415: agreed. fixed.

core/crypto/nike/hybrid/no_nobs_csidh.go
david415: This is a convenience to allow us to compile our mix server and dirauth server on that ppc64le hardware. These are not even used by the mix server or the dirauth server... but the build would fail.

david415: I'm fine with removing it but then mixserver and dirauth won't build on ppc64le.

core/sphinx/sphinx_benchmark_test.go
david415: fixed

core/sphinx/sphinx_ctidh_test.go
david415: that test will not run by default unless you specify the ctidh tag... however just that tag alone is not sufficient.. there's a bunch of other env vars you need to specify in order to make it compile with CTIDH. it's a pain and I'm planning on putting all that stuff in a makefile at the root of the monorepo soon.

//go:build ctidh

core/wire/kem.go
david415: I want to export it so that we can thoroughly test it from inside of the core/crypto/kem module:
crypto/kem/schemes/kem_test.go

david415: but actually we don't need to define the kem scheme inside core/wire... probably better to define it in core/crypto/kem/hybrid

david415: hmm or maybe we do want the semantics of "run these kem tests on whatever kem the core/wire protocol uses"

david415: okay i've "fixed" it.

core/crypto/kem/adapter/kem_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/katzenpost/katzenpost/core/crypto/rand"
+)
+
+func TestNikeToKemAdapter(t *testing.T) {
+	ecdhNike := ecdh.NewEcdhNike(rand.Reader)
+	s := FromNIKE(ecdhNike)
+
+	t.Logf("hello my name is %s", s.Name())
+
+	pubkey1, privkey1, err := s.GenerateKeyPair()
```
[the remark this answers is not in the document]
david415: At your previous request, i added a deterministic KEM test for our hybrid KEM (which core/wire uses); Should I also add another such test here as well? Technically the test vectors for the hybrid which uses this adapter should cover the behavior of this adapter.

core/crypto/kem/hybrid/hybrid.go
david415: fixed!

docs/specs/kemsphinx.rst
david415: Yes, exactly. cct1 and cct2 are KEM ciphertexts. Although one of them is from the NIKE to KEM adapter.

docs/specs/kemsphinx.rst
david415: we aren't allowed to use snake case in golang. it's camel case. however i prefer to write my specs with snake case because it's easier to read.

core/crypto/kem/adapter/kem.go
david415: no.

david415: i totally agree snake case is better. but i think it's actually really important to follow the conventions of a programming language. the golang compiler will build if we use snake case so this isn't a technical argument. but i am trying to make a social argument for following the convention.

docs/specs/kemsphinx.rst
threebithacker: Cool, we should put that in the spec so implementers can do it clean-room without having to go scouring the implementation to figure out what `cct` and `ss` are supposed to be.

docs/specs/kemsphinx.rst
threebithacker: My suggestion is to fix the spec like this:
```diff
 func ENCAPSULATE(their_pub_key publickey) ([]byte, []byte) {
         my_priv_key, my_pub_Key = GEN_KEYPAIR(RNG)
         ss = DH(my_priv_key, their_pub_key)
-        ss2 = H(ss || my_pubkey || their_pubkey)
+        ss2 = H(ss || their_pubkey || my_pubkey)
         return my_pub_key, ss2
```

core/crypto/kem/adapter/kem.go
threebithacker: I don't care about snake case or camel case. I care about `pubkey1` and `pubkey2` being nondescript and confusing here compared to `theirPubkey, myPubkey`. Or in this case since it's used in both functions, `encapsulatorPubkey`, `decapsulatorPubkey` probably.

core/crypto/kem/hybrid/hybrid.go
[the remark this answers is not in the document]
threebithacker: We now have `EncapsulationSeedSize() = sch.first.EncapsulationSeedSize() + sch.second.EncapsulationSeedSize()`
so we can delete the XOF and hash stuff and just call:
```go
	ct1, ss1, err := sch.first.EncapsulateDeterministically(pub.first, seed[ : sch.first.EncapsulationSeedSize() ])
// ...
	ct2, ss2, err := sch.second.EncapsulateDeterministically(pub.second, seed[ sch.first.EncapsulationSeedSize() : ])
```

core/sphinx/sphinx_ctidh_test.go
threebithacker: Cool, fair enough

docs/specs/kemsphinx.rst
threebithacker: Sorry, I was reading the old commit. New text in `2.2` does explain it, thanks!

threebithacker (changes requested):

core/crypto/nike/hybrid/no_nobs_csidh.go
threebithacker: Why is `NOBS_CSIDHX25519` required to build `mixserver` and `dirauth` on `ppc64le` when the implementation isn't required to work, or called?

core/crypto/kem/adapter/kem_test.go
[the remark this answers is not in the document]
david415: core/crypto/kem/schemes/kem_test.go is where the test vector deterministic test lives for testing our hybrid KEM which we use for Noise. but i'll also put a note in this source file.

docs/specs/kemsphinx.rst
```diff
@@ -30,7 +32,73 @@ has two functions:
...
+-----------------------
+
+We easily achieve our IND-CCA2 security by means of hashing
+together the DH shared secret along with both of the public keys:
+
+```
+func ENCAPSULATE(their_pub_key publickey) ([]byte, []byte) {
+        my_priv_key, my_pub_Key = GEN_KEYPAIR(RNG)
+        ss = DH(my_priv_key, their_pub_key)
+        ss2 = H(ss || their_pubkey || my_pubkey)
```
threebithacker, line 61:
```suggestion
        ss2 = H(ss || their_pub_key || my_pubkey)
```
Sorry man, I put `their_pubkey` but you have `their_pub_key` in the other lines

docs/specs/kemsphinx.rst
threebithacker: Thanks!

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,267 @@
...
+// Scheme is an adapter for nike.Scheme to kem.Scheme.
+// See docs/specs/kemsphinx.rst for some design notes
+// on this NIKE to KEM adapter.
+type Scheme struct {
+	nike nike.Scheme
+}
+
+var _ kem.Scheme = (*Scheme)(nil)
+var _ kem.PublicKey = (*PublicKey)(nil)
+var _ kem.PrivateKey = (*PrivateKey)(nil)
```
threebithacker, line 77: very nice with static type assertions!

threebithacker (changes requested):

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,267 @@
...
+func (a *Scheme) PrivateKeySize() int {
+	return a.nike.PrivateKeySize()
+}
+
+// Size of packed public keys.
+func (a *Scheme) PublicKeySize() int {
+	return a.nike.PublicKeySize()
+}
+
+// DeriveKeyPair deterministicallly derives a pair of keys from a seed.
```
threebithacker, line 209:
```suggestion
// DeriveKeyPair deterministically derives a pair of keys from a seed.
```

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,267 @@
...
+	if len(seed) != a.SeedSize() {
+		panic(fmt.Errorf("%s: provided seed of length %d is != to correct seed size of %d", kem.ErrSeedSize, len(seed), a.SeedSize()))
+	}
+	h, err := blake2b.NewXOF(0, nil)
+	if err != nil {
+		panic(err)
+	}
+
+	seedHash := blake2b.Sum256(seed)
+	_, _ = h.Write(seedHash[:])
```
threebithacker, line 222: What are the return values ignored here (`_, _ = h.Write`) - should it have a `panic()` on error? 

I'm assuming `NewXOF` is required because `a.nike.GenerateKeyPairFromEntropy` expects a hash state as input?

threebithacker (changes requested):

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,267 @@
...
+func (a *Scheme) PublicKeySize() int {
+	return a.nike.PublicKeySize()
+}
+
+// DeriveKeyPair deterministicallly derives a pair of keys from a seed.
+// Panics if the length of seed is not equal to the value returned by
+// SeedSize.
+func (a *Scheme) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
+	if len(seed) != a.SeedSize() {
+		panic(fmt.Errorf("%s: provided seed of length %d is != to correct seed size of %d", kem.ErrSeedSize, len(seed), a.SeedSize()))
```
threebithacker, line 214:
```suggestion
		panic(fmt.Errorf("%s: provided len(seed) %d != a.SeedSize() %d", kem.ErrSeedSize, len(seed), a.SeedSize()))
```
Mentioning the parameter name and the constant it's supposed to match seems like an easy life quality improvement for whoever triggers this assert here :)

threebithacker (changes requested):

core/crypto/kem/adapter/kem.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+func hashSharedSecretWithPublicKeys(publicKeySize int, ss []byte, pubkey1 []byte, pubkey2 []byte) []byte {
+	var h blake2b.XOF
+	var err error
+	if len(ss) != 32 {
+		sum := blake2b.Sum256(ss)
+		h, err = blake2b.NewXOF(uint32(publicKeySize), sum[:])
+	} else {
+		h, err = blake2b.NewXOF(uint32(publicKeySize), ss)
```
threebithacker, line 125: `publicKeySize` here should semantically come from `a.SharedKeySize()` (they're equivalent in this file) I don't think we need to receive it as a parameter in this function, but it might be useful as a utility function elsewhere or something. I don't know, I'd learn towards either:
1. Just hardcoding `blake2b.NewXOF(uint32(a.SharedKeySize()), ss)` here
2. renaming the function parameter to `sharedKeySize` and changing the call sites to use `a.SharedKeySize()` instead of `a.PublicKeySize()`

threebithacker (changes requested):

core/wire/session.go
```diff
@@ -184,22 +184,23 @@ func (s *Session) handshake() error {
 	defer handshake.Reset()
 	var (
 		prologueLen = 1
+		keyLen      = 32
```
threebithacker, line 187: Which `key` is this the length of? Is there a constant in `nyquist` we should be referencing here?

threebithacker (changes requested):

core/wire/session.go
david415: ah yeah Nyquest has a constant for that... i will fix

core/crypto/kem/adapter/kem.go
david415: So... Golang is a messy language. In my humble opinion there should be a separate interface for writes or reads that cannot fail or whose operation will never partially succeed... and so here's what golang io.Writer is really about:
https://pkg.go.dev/io#Writer

The first return value is length written. If it's not equal to the desired length written then we should panic; that having been said it probably cannot partially fail in this way because it's just a numerical calculation. anyway I've fixed this.

core/crypto/kem/adapter/kem.go
david415: fixed

core/crypto/kem/adapter/kem.go
david415: fixed

core/crypto/nike/hybrid/no_nobs_csidh.go
threebithacker: Seems to me it would be more appropriate to have `geo.go` parametrically compiled then, to only reference libraries available on the target platform?

threebithacker: Seems to me it would be more appropriate to have `geo.go` parametrically compiled then, to only reference libraries available on the target platform?

threebithacker (approved):
Looks good to me.

PR #260, opened 2023-05-04
catshadow: add client method DoubleRatchetPayloadLength

david415:
This simple PR adds a catshadow client method that helps out with the upgrade


leif (approved):

leif:
this is used here https://github.com/katzenpost/katzen/pull/32

Issue #263, opened 2023-05-05
cbor plugin worker deadlocks if plugin returns error and no response


mixmasala:
I noticed that if a kaetzchen plugin returns an error rather than a cborplugin.Response, the cborplugin worker will deadlock. This might explain why some services like PANDA are unreliable.

See:
https://github.com/katzenpost/katzenpost/blob/731353c7f1065871afcd4e4c6da10696a5c82c83/server/internal/provider/kaetzchen/cbor_plugins.go#L138

https://github.com/katzenpost/katzenpost/blob/731353c7f1065871afcd4e4c6da10696a5c82c83/panda/server/cmd/panda_server/main.go#L120

https://github.com/katzenpost/katzenpost/blob/731353c7f1065871afcd4e4c6da10696a5c82c83/memspool/server/cmd/memspool/main.go#L116

https://github.com/katzenpost/katzenpost/blob/731353c7f1065871afcd4e4c6da10696a5c82c83/server/cborplugin/server.go#L76

mixmasala:
to fix this, the cborplugin worker should write a command respresenting the error to the WriteChan

mixmasala:
time spent: 2 hours

mixmasala:
resolved

PR #264, opened 2023-05-06
server: cborplugin worker must always write a response

mixmasala:
this fixes a deadlock issue reported in #263


david415 (approved):
lgtm

PR #269, opened 2023-05-13
Fix csidh ppc64le noop

mixmasala:
this disables csidh on ppc64le; which is necessary because some of our nodes run on ppc64le but don't run clients.


ioerror:
I am using branch to build katzenpost on anarres (power9/ppc64le).

ioerror (commented):
This branch fixes the build of v0.0.23 katzenpost on POWER9/ppc64le but I have not yet seen that build produce a consensus with that binary. It isn't yet clear if this solves one problem but creates another or if it's entirely unrelated. It appears unrelated but I'm not comfortable with approving this until anarres works as expected.

PR #272, opened 2023-05-13
Fix pki client log line

david415:


leif:
this fixes the problem where the pki client would log `GetConsensus(156425) (attempt 0, err=<nil>, res=Conflict)` when it meant `res=Gone` (they're both `2`).

PR #277, opened 2023-05-14
docs: WIP: update PKI documentation

david415:
These are good changes to the PKI spec which i thought we had merged long ago. We should merge this.


ioerror (changes requested):
LGTM; the CI should pass and then this is ready for a merge as it correctly reflects the current PKI situation as I understand it.

docs/specs/pki.rst
```diff
@@ -183,8 +184,8 @@ Katzenpost Mix Network Public Key Infrastructure Specification
 
    Thus, at any time, keys for all Mixes for the Nth through N + 2nd
    epoch will be available, allowing for a maximum round trip (forward
-   message + SURB) delay + transit time of 6 hours. SURB lifetime is
-   limited to a few hours because of the key rotation epoch, however
+   message + SURB) delay + transit time of 40 minutes. SURB lifetime is
+   limited to a single epoch because of the key rotation epoch, however
```
threebithacker, line 188: If there's a grace period then "single epoch" sounds wrong to me, but we could make it "at most two epochs" or mention the grace period here. Oh well.

threebithacker (approved):

docs/specs/pki.rst
mixmasala: To clarify, here is the function that implements this behavior
```
func (w *Worker) doUnwrap(pkt *packet.Packet) error {
        const gracePeriod = 2 * time.Minute

        // Figure out the candidate mix private keys for this packet.
        keys := make([]*mixkey.MixKey, 0, 2)
        epoch, elapsed, till := epochtime.Now()
        k, ok := w.mixKeys[epoch]
        if !ok || k == nil {
                // There always will be a key for the current epoch, since
                // key generation happens multiple epochs in advance.
                return fmt.Errorf("crypto: No key for epoch %v", epoch)
        }
        keys = append(keys, k)

        // At certain times, this needs to also look at the previous
        // or next epoch(s) keys, if they exist.
        if elapsed < gracePeriod {
                // Less than gracePeriod into the current epoch, the previous
                // epoch's key should also be accepted.
                k, ok = w.mixKeys[epoch-1]
        } else if till < gracePeriod {
                // Less than gracePeriod to the next epoch, the next epoch's
                // key should also be accepted.
                k, ok = w.mixKeys[epoch+1]
        } else {
                // Only one key to use.
                k = nil
                ok = false
        }
        if ok && k != nil {
                // Not having other keys is fine, regardless of if we are
                // in the grace period, if a packet happens to get dropped,
                // oh well.
                keys = append(keys, k)
        }
```
the grace period accepts keys from the previous epoch for only 2 minutes straddling the transition.

mixmasala: I've updated the spec to mention the grace period, and opened #325

mixmasala: I've updated the spec to mention the grace period and opened a ticket regarding the grace period #325

PR #281, opened 2023-05-18
docker: Less aggressive default decoy traffic

david415:
The current default docker decoy traffic setting is so aggressive that my laptop freezes up for a few seconds at a time while I'm trying to type into my text editor and I am very frustrated. Let's merge this.


david415:
@mixmasala please merge

ioerror (approved):
LGTM

PR #283, opened 2023-05-19
Log sphinx geometry on startup and error

david415:


ioerror:
If the CI passes, I'll review this branch.

PR #284, opened 2023-05-20
Add reliable streams

mixmasala:
This adds the map storage system and stream transport


map/server/server.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+		if _, err := tx.CreateBucketIfNotExists([]byte(gcBucket)); err != nil {
+			return err
+		}
+		return nil
+	}); err != nil {
+		m.db.Close()
+		log.Errorf("%s", err)
+		panic(err)
+		return nil, err
```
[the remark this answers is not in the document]
mixmasala: fixed

sockatz/socks5/rfc1929.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// Pluggable transports use the username/password field to pass
+	// per-connection arguments.  The fields contain ASCII strings that
+	// are combined and then parsed into key/value pairs.
+	argStr := string(uname)
+	if !(plen == 1 && passwd[0] == 0x00) {
+		// tor will set the password to 'NUL', if the field doesn't contain any
+		// actual argument data.
+		argStr += string(passwd)
```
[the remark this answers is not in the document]
mixmasala: excuse me?

stream/buffered_test.go
```diff
@@ -0,0 +1,172 @@
...
+	return a, b
+}
+
+func NewMockTransport() Transport {
+	m := &mockTransport{}
+	m.l = new(sync.Mutex)
+	m.l.Lock()
+	defer m.l.Unlock()
+	m.data = make(map[string][]byte)
+	return m
```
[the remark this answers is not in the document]
mixmasala: idk this compiles without error.

map/server/server_test.go
```diff
@@ -0,0 +1,141 @@
...
+	var msgID common.MessageID
+	_, err = rand.Reader.Read(msgID[:])
+	require.NoError(err)
+	payload := []byte("hola")
+
+	err = m.Put(msgID, payload)
+	require.NoError(err)
+
+	// read data from key
+	data, err := m.Get(msgID)
```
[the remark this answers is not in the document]
mixmasala: fixed

map/server/server.go
```diff
@@ -0,0 +1,251 @@
...
+	m := &Map{
+		log:     log,
+		mapSize: mapSize,
+		gcSize:  gcSize,
+	}
+	db, err := bolt.Open(fileStore, 0600, nil)
+	if err != nil {
+		log.Errorf("%s", err)
+		panic(err)
+		return nil, err
```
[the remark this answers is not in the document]
mixmasala: fixed

map/server/server.go
```diff
@@ -0,0 +1,251 @@
...
+			}
+			binary.BigEndian.PutUint64(b[:], i)
+			gcbkt, err = bkt.CreateBucketIfNotExists(b[:])
+			if err != nil {
+				return err
+			}
+		}
+
+		// store msgID in gcBucket
+		gcbkt.Put(msgID[:], []byte{0x1})
```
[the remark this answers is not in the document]
mixmasala: fixed

map/server/server.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	m.Go(m.worker)
+	return m, nil
+}
+
+func (m *Map) OnCommand(cmd cborplugin.Command) (cborplugin.Command, error) {
+	switch r := cmd.(type) {
+	case *cborplugin.Request:
+		if !r.HasSURB {
+			return nil, errors.New("No SURB, cannot reply")
```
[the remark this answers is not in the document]
mixmasala: fixed

threebithacker (changes requested):

stream/stream_test.go
```diff
@@ -0,0 +1,355 @@
...
+				s.Halt()
+				return
+			}
+			b := make([]byte, len(msg))
+			// Read() data until we have received the message
+			for readOff := 0; readOff < len(msg); {
+				n, err := s.Read(b[readOff:])
+				if err != nil {
+					t.Logf("read %d, total %d", n, readOff)
+					if readOff != len(msg) {
```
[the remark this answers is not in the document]
mixmasala: I believe Read() may return io.EOF and a nonzero n.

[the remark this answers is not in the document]
mixmasala: I've changed the condition to check for the correct length on error and fail the test accordingly.

sockatz/socks5/rfc1929.go
```diff
@@ -0,0 +1,91 @@
...
+	// Read the username.
+	var ulen byte
+	if ulen, err = req.readByte(); err != nil {
+		sendErrResp()
+		return
+	} else if ulen < 1 {
+		sendErrResp()
+		return fmt.Errorf("username with 0 length")
+	}
+	var uname []byte
```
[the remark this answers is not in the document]
mixmasala: removed from tree

sockatz/socks5/rfc1929.go
```diff
@@ -0,0 +1,91 @@
...
+	// Read the password.
+	var plen byte
+	if plen, err = req.readByte(); err != nil {
+		sendErrResp()
+		return
+	} else if plen < 1 {
+		sendErrResp()
+		return fmt.Errorf("password with 0 length")
+	}
+	var passwd []byte
```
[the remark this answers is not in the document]
mixmasala: removed from tree

stream/stream_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+	// worker B
+	go func() {
+		s, err := DialDuplex(session, "", addr.String())
+		require.NoError(err)
+		for {
+			msg, ok := <-sidechannel
+			// channel was closed by writer, we're done
+			if !ok {
+				// verify that EOF is (eventuallY) returned on a Read after Stream.Close()
```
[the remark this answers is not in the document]
mixmasala: resolved

stream/buffered_test.go
```diff
@@ -0,0 +1,172 @@
...
+	a := newStream(t)
+	a.log = logBackend.GetLogger(fmt.Sprintf("Stream %x", &a))
+	addr := &StreamAddr{address: generate()}
+	a.keyAsListener(addr)
+	b := newStream(t)
+	b.log = logBackend.GetLogger(fmt.Sprintf("Stream %x", &b))
+	b.keyAsDialer(addr)
+
+	if a == nil || b == nil {
+		panic("wtf")
```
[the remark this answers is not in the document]
mixmasala: &mockTransport implements Transport (type Transport is an interface, so it is already a pointer), and the value returned is a pointer to mockTransport.

see https://github.com/katzenpost/katzenpost/blob/1bec21a5cefe9c94049c034631594254d488850f/stream/stream.go#L866
https://github.com/katzenpost/katzenpost/blob/1bec21a5cefe9c94049c034631594254d488850f/map/client/client.go#L165

stream/buffered_test.go
```diff
@@ -0,0 +1,172 @@
...
+type mockTransport struct {
+	l    *sync.Mutex
+	data map[string][]byte
+}
+
+// newStreams returns an initialized pair of Streams
+func newStreams(t Transport) (*Stream, *Stream) {
+
+	a := newStream(t)
+	a.log = logBackend.GetLogger(fmt.Sprintf("Stream %x", &a))
```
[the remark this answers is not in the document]
mixmasala: sure. applied.

stream/buffered_test.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}()
+	go func() {
+		addr := make([]byte, 8)
+		for i := 0; i < 4096; i++ {
+			binary.BigEndian.PutUint64(addr, uint64(i))
+			garbage.Put(addr, randPayload())
+		}
+	}()
+
```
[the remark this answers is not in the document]
mixmasala: yes. nice catch. fixed.

stream/buffered_test.go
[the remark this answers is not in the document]
mixmasala: Transport is an interface type, and mockTransport implements it

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+// rwMap implements ReadWriteTranport using a ReadWriteCap and reference to map client
+// The idea is that several streams can share map/client and client/session but
+// have different capabilities for the underlying Streams transported by map/client
+type rwMap struct {
+	c     *Client
+	rwCap common.ReadWriteCap
+}
+
+// Get implements RWClient.Get
+func (r *rwMap) Get(addr []byte) ([]byte, error) {
```
david415, line 192: ideally we'd be able to test the full functionality of a capability without any client IO

map/server/server.go
```diff
@@ -0,0 +1,252 @@
...
+		// verify v Signs the publickey bytes
+		return v.Verify(req.Signature, req.ID.Bytes())
+	} else {
+		v := req.ID.WritePk()
+		// verify v Signs the payload bytes
+		return v.Verify(req.Signature, req.Payload)
+	}
+}
+
+func (m *Map) RegisterConsumer(svr *cborplugin.Server) {
```
david415, line 250: not used. remove

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	slot := int(binary.LittleEndian.Uint64(ID[:8])) % len(descs)
+	// sort the descs and return the chosen one
+	desc := deterministicSelect(descs, slot)
+	return &mapStorage{name: desc.Name, provider: desc.Provider, id: ID}, nil
+}
+
+// Put places a value into the store
+func (c *Client) Put(ID common.MessageID, signature, payload []byte) error {
+	if !ID.WritePk().Verify(signature, payload) {
```
david415, line 105: Client's Put method here is mixing IO with cryptography which would normally be totally acceptable if the core protocol actually had the call to the cryptographic verification. as it stands now, this abstraction is difficult to test.

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+// PayloadSize returns the size of the user payload
+func (c *Client) PayloadSize() int {
+	geo := c.Session.SphinxGeometry()
+	return geo.UserForwardPayloadLength - cborFrameOverhead - eddsa.SignatureSize
+}
+
+// Get requests ID from the chosen storage node and returns a payload or error
+func (c *Client) Get(ID common.MessageID, signature []byte) ([]byte, error) {
+
+	if !ID.ReadPk().Verify(signature, ID[:]) {
```
david415, line 135: here is another example of mixing IO with cryptography which would in some contexts be totally fine however in this case that's not a cryptography operation that is woven into the core cryptography module.

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+
+// Get implements RWClient.Get
+func (r *rwMap) Get(addr []byte) ([]byte, error) {
+	i := r.rwCap.Addr(addr)
+	k := r.rwCap.Read(addr)
+	return r.c.Get(i, k.Sign(i.Bytes()))
+}
+
+// Put implements RWClient.Put
+func (r *rwMap) Put(addr []byte, payload []byte) error {
```
david415, line 199: here's another example of a method that needs to be rewritten so that we can test the math separately from the io

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// roMap implements ReadOnlyTranport using a ReadOnlyCap and reference to map client
+type roMap struct {
+	c     *Client
+	roCap common.ReadOnlyCap
+}
+
+// Get implements ROClient.Get
+func (r *roMap) Get(addr []byte) ([]byte, error) {
```
david415, line 217: here's another example of a method that needs to be rewritten so that we can test the math separately from the io

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// woMap implements WOClient
+type woMap struct {
+	c     *Client
+	woCap common.WriteOnlyCap
+}
+
+// Put implements WOClient.Put
+func (w *woMap) Put(addr []byte, payload []byte) error {
```
david415, line 235: here's another example of a method that needs to be rewritten so that we can test the math separately from the io

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+func WriteOnly(c *Client, woCap common.WriteOnlyCap) WOClient {
+	m := new(woMap)
+	m.c = c
+	m.woCap = woCap
+	return m
+}
+
+// create a duplex using a shared secret
+func DuplexFromSeed(c *Client, initiator bool, secret []byte) RWClient {
+	salt := []byte("duplex initialized from seed is not for multi-party use")
```
david415, line 272: here's another example of a method that needs to be rewritten so that we can test the math separately from the io

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+// RWClient with different read/write root capabilities so that a pair
+// of clients may use the capabilities to communicate unidirectionally
+type duplex struct {
+	ro ROClient // used to read data to client
+	wo WOClient // used to send data to peer
+}
+
+// Put implements RWClient.Put
+func (s *duplex) Put(addr []byte, payload []byte) error {
+	return s.wo.Put(addr, payload)
```
david415, line 311: here's another example of a method that needs to be rewritten so that we can test the math separately from the io

map/common/cap.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// ReadOnlyCap describes a Capability that has Write capability only.
+type WriteOnlyCap interface {
+	Cap
+	Write(addr []byte) *eddsa.BlindedPrivateKey
+}
+
+// PCap holds CapPk and Addr method
+type PCap struct {
```
david415, line 79: PCap is a name that doesn't tell the developer anything. Let's pick better names for things.

stream/buffered.go
```diff
@@ -0,0 +1,125 @@
...
+	"errors"
+	"io"
+	"sync"
+
+	"github.com/fxamacker/cbor/v2"
+	"github.com/katzenpost/katzenpost/core/worker"
+)
+
+// BufferedStream holds a Stream and Buffer of bytes that are used with cbor.Decoder
+type BufferedStream struct {
```
david415, line 14: where is this used?

stream/stream.go
```diff
@@ -0,0 +1,1013 @@
...
+	// StreamData indicates that this is a data carrying Frame in a Stream
+	StreamData
+	// StreamEnd indicates that this is the last Frame in a Stream
+	StreamEnd
+)
+
+// Frame is the container for Stream payloads and contains Stream metadata
+// that indicates whether the Frame is the first, last, or an intermediary
+// block. This
+type Frame struct {
```
david415, line 60: why are some field exported and others not? perhaps all the fields can be unexported?

map/client/client.go
```diff
@@ -0,0 +1,335 @@
...
+// WOClient only has Put
+type WOClient interface {
+	Put(addr []byte, payload []byte) error
+	PayloadSize() int
+}
+
+// rwMap implements ReadWriteTranport using a ReadWriteCap and reference to map client
+// The idea is that several streams can share map/client and client/session but
+// have different capabilities for the underlying Streams transported by map/client
+type rwMap struct {
```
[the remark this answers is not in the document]
mixmasala: See https://github.com/katzenpost/katzenpost/blob/add_reliable_streams/map/common/cap.go for the capability system.

map/client/client.go
mixmasala: I agree that https://github.com/katzenpost/katzenpost/blob/add_reliable_streams/map/common/cap_test.go is incomplete and will add additional tests for each capability type.

map/common/cap_test.go
```diff
@@ -0,0 +1,39 @@
...
+package common
+
+import (
+	"github.com/katzenpost/katzenpost/core/crypto/eddsa"
+	"github.com/katzenpost/katzenpost/core/crypto/rand"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestCreateRWCap(t *testing.T) {
```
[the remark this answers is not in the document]
mixmasala: Agree that more tests are needed for the capability system.

map/client/client.go
mixmasala: these abstractions implement interfaces and do not test the capability system.

map/common/cap.go
mixmasala: PCap type provides method Addr and the capability public key for each type that inherits from it. Do you have a suggestion for a better name? PublicCapability? BaseCapability?

stream/buffered.go
mixmasala: BufferedStream is used when using a cbor.Decoder with a Stream: if a CBOR object is incompletely decoded before a Stream is Halt()'d then the contents read from the stream must be preserved in order to resume deserializing CBOR from the Stream in the future. Katzen uses this here:
https://github.com/katzenpost/katzen/blob/3451457574303f0604cc00ec7a0b722653e38c18/katzen.go#L90

mixmasala: https://github.com/katzenpost/katzen/blob/3451457574303f0604cc00ec7a0b722653e38c18/katzen.go#L212

client/send.go
```diff
@@ -244,8 +245,9 @@ func (s *Session) BlockingSendUnreliableMessage(recipient, provider string, mess
 	select {
 	case reply := <-replyWaitChan:
 		return reply, nil
-	// these timeouts are often far too aggressive
-	case <-time.After(sentMessage.ReplyETA + cConstants.RoundTripTimeSlop):
+	case <-s.HaltCh():
+		return nil, ErrHalted
+	case <-time.After(sentMessage.ReplyETA + sentMessage.ReplyETA>>1):
```
david415, line 250: Using >> or << is far less readable than / or *. It also doesn't do what you think unless they are unsigned ints... furthermore regardless of the operator used here for division, there needs to be parentheses so that the reviewer doesn't need to go look up the golang specification and read the operator precedence ordering which is:

Precedence    Operator
    5             *  /  %  <<  >>  &  &^
    4             +  -  |  ^
    3             ==  !=  <  <=  >  >=
    2             &&
    1             ||

mixmasala: d04ecb98f04d5cca08da1d2a9e00efe07679b1e2 f7427a08fa76fad1243d5fe9c76df1c8ef413262 change the timeout behavior to use the RoundTripTimeSlop, reduces RoundTripTimeSlop to 4 seconds, and changes the clients garbage collector behavior so that kaetzchen responders can respond within 2 epochs

david415:
```
client_test.go:99: 
        	Error Trace:	/go/katzenpost/map/client/client_test.go:99
        	Error:      	Not equal: 
        	            	expected: []byte{0x67, 0x6f, 0x6f, 0x64, 0x62, 0x79, 0x65, 0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64}
        	            	actual  : []byte{0x68, 0x65, 0x6c, 0x6c, 0x6f, 0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	-([]uint8) (len=13) {
        	            	- 00000000  67 6f 6f 64 62 79 65 20  77 6f 72 6c 64           |goodbye world|
        	            	+([]uint8) (len=11) {
        	            	+ 00000000  68 65 6c 6c 6f 20 77 6f  72 6c 64                 |hello world|
        	            	 }
        	Test:       	TestCreateMap
--- FAIL: TestCreateMap (9.16s)

```

map/client/client.go
david415: Where is the test for this function?

david415: `DuplexFromSeed` needs a unit test.

david415 (changes requested):
`DuplexFromSeed` needs a test

map/client/client.go
david415: ideally we would separate it away the pure math from functions that take a client reference as an argument.

mixmasala:
>     * [ ]  buffered.go: unused??? also using empty interfaces in CBORDecode call chain
> 

BufferedStream will try to decode cbor into the instance passed to it, so the user can use Stream with any serializable types.
>     * [ ]  stream.Frame: why is id private but everything else Public?
I think this might have been a bug because Frames in the Wack map (Waiting Ack) would be saved/restored with Frame.id = 0

>     * [ ]  stream.Stream: why some fields internal and external

everything needed to serialize/deserialize the state of a Stream are external, except internal fields that shouldn't be exported - such as mutex, channels, logger instances, etc

david415 (approved):

PR #285, opened 2023-05-20
Add quic listeners

mixmasala:
This PR adds support for QUIC transport using https://github.com/quic-go/quic-go and deprecates AltAddresses.


authority/voting/server/server.go
```diff
@@ -307,14 +364,47 @@ func New(cfg *config.Config) (*Server, error) {
...
 	// Start up the listeners.
 	for _, v := range s.cfg.Server.Addresses {
-		l, err := net.Listen("tcp", v)
-		if err != nil {
-			s.log.Errorf("Failed to start listener '%v': %v", v, err)
-			continue
+		// parse the Address line as a URL
+		u, err := url.Parse(v)
+		if err == nil {
+			switch u.Scheme {
```
leif, line 370: it looks like you're only getting `tcp://` here, not `tcp4://` and `tcp6://`

leif:
`http` doesn't seem like the right name for the scheme, since it is really just QUIC. should it be `quic://` instead?

replying to the commit message in [0ebb780](https://github.com/katzenpost/katzenpost/pull/285/commits/0ebb7806117486ca658aaaca7602757ff1666c92):

> There are some outstanding details such as:
> 
> * Where or how should TLS certificates be provided for the Authority?
>  How are these certificates verified in the quic-go library? Can we put
>  the certificate fingerprint in the client configuration or do we have
>  to embed the entire certificate in the configuration?
>
> * For now the QUIC transport TLS certificate verification is disabled.

This isn't a regression per se since QUIC is just replacing TCP here (we're still using the PQ noise transport over QUIC) but it would obviously be preferable to verify the QUIC keys too.

> * The unit tests for the client and server use a net.Pipe and mock out
>  the transport details; there need to be new tests written to verify
>  client and server configured with multiple transport types.

:+1: 

> * Katzenpost's proxy support is implemented by replacing the
>  defaultDialer type - which is mostly used with Tor via SOCKS5 - and
>  this doesn't support UDP - I override the dialer type but a cleaner
>  solution would be preferable because it would break the users proxy
>  choices.

Isn't it be easy to just ignore the quic transport when UpstreamProxy is in use?

> * Because the QUIC supports multiple Streams inside of a Connection, the
>  go interface types aren't an exact match. The common.QuicConn and
>  common.QuicListener wrapper types implement net.Conn and net.Listener
>  for a single Stream.

:shrug: we only want to make one stream per connection anyway, right?

> * We'll probably actually want to put the QUIC wrapper stuff in a path
> common to both the Authority and Server components

:+1:

authority/voting/server/server.go
mixmasala: yeah, I think we should get rid of tcp4 and tcp6 since there is no such thing ;-)

mixmasala:
> `http` doesn't seem like the right name for the scheme, since it is really just QUIC. should it be `quic://` instead?
> 

I had a similar thought. It's also silly that http3 has the same URL scheme so browsers have to try both TCP and UDP.
I think changing to quic:// is a good idea.

> replying to the commit message in [0ebb780](https://github.com/katzenpost/katzenpost/pull/285/commits/0ebb7806117486ca658aaaca7602757ff1666c92):
> 
> > There are some outstanding details such as:
> > 
> > * Where or how should TLS certificates be provided for the Authority?
> >   How are these certificates verified in the quic-go library? Can we put
> >   the certificate fingerprint in the client configuration or do we have
> >   to embed the entire certificate in the configuration?
> > * For now the QUIC transport TLS certificate verification is disabled.
> 
> This isn't a regression per se since QUIC is just replacing TCP here (we're still using the PQ noise transport over QUIC) but it would obviously be preferable to verify the QUIC keys too.

It's still an improvement over TCP where an adversary who wants to reset the connection now needs to MITM TLS, but it could be improved, where it would be ideal if we could use TLS-PSK. Other options are to use the 
> 
> > * The unit tests for the client and server use a net.Pipe and mock out
> >   the transport details; there need to be new tests written to verify
> >   client and server configured with multiple transport types.
> 

I modified our end-to-end tests to create a mix of tcp and quic listener, as well as used the end-to-end tests with all components (mixes, authorities, and clients) using quic only. This PR uses a mix of quic and tcp.

> +1
> 
> > * Katzenpost's proxy support is implemented by replacing the
> >   defaultDialer type - which is mostly used with Tor via SOCKS5 - and
> >   this doesn't support UDP - I override the dialer type but a cleaner
> >   solution would be preferable because it would break the users proxy
> >   choices.
> 
> Isn't it be easy to just ignore the quic transport when UpstreamProxy is in use?

Ok, that's a good idea for a temporary solution - the correct solution is to implement 
> 
> > * Because the QUIC supports multiple Streams inside of a Connection, the
> >   go interface types aren't an exact match. The common.QuicConn and
> >   common.QuicListener wrapper types implement net.Conn and net.Listener
> >   for a single Stream.
> 
> shrug we only want to make one stream per connection anyway, right?

Well, we currently only allow clients to make one connection in order to implement client rate limiting.
> 
> > * We'll probably actually want to put the QUIC wrapper stuff in a path
> >   common to both the Authority and Server components
> 
That comment was addressed, see path http
> +1

http/common/quic.go
```diff
@@ -0,0 +1,180 @@
...
+	"net/url"
+	"time"
+
+	"github.com/quic-go/quic-go"
+	"github.com/quic-go/quic-go/http3"
+)
+
+// QuicConn wraps a conn and a single stream and implements net.Conn
+type QuicConn struct {
+	Stream quic.Stream
```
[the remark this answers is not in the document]
mixmasala: this seems like a bug because sonatype doesn't check the imported package name.

mixmasala:
Fixed a few issues and extended our docker/Makefile to support running mixnets with specific transports only.

authority/voting/server/server.go
mixmasala: updated to catch tcp4 and tcp6 as well.

.github/workflows/go.yml
```diff
@@ -19,7 +20,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Generate mixnet configuration
-        run: cd genconfig && go build && ./genconfig -v -b /conf -o ../docker/voting_mixnet/
+        run: cd genconfig && go build && ./genconfig -v -onlyTransports ${{ matrix.transport }} -b /conf -o ../docker/voting_mixnet/
```
david415, line 23: does that make it only use one transport protocol for the entire mixnet? e.g. QUIC?

.github/workflows/go.yml
```diff
@@ -35,6 +36,7 @@ jobs:
       matrix:
         go-version: [1.19.x]
         os: [ubuntu-22.04]
+        transport: [ "", "tcp", "quic" ]
```
david415, line 39: it tests it separately for each protocol?
what happens with the "" protocol?

authority/voting/server/server.go
```diff
@@ -138,6 +142,35 @@ func (s *Server) listenWorker(l net.Listener) {
...
+			s.log.Errorf("accept failure: %v", err)
+			return
+		}
+		s.Add(1)
+		s.onConn(conn)
+	}
+	// NOTREACHED
+}
+
+func (s *Server) listenWSWorker(l net.Listener) {
```
david415, line 165: does this work? if not we should remove it

authority/voting/server/wire_handler.go
```diff
@@ -37,8 +38,17 @@ func (s *Server) onConn(conn net.Conn) {
 	)
 
 	rAddr := conn.RemoteAddr()
-	s.log.Debugf("Accepted new connection: %v", rAddr)
+	var proto string
+	switch conn.(type) {
+	case *net.TCPConn:
+		proto = "tcp"
```
david415, line 44: does there need to be a tcp6 case here? i guess not. there's only a net.TCPConn type.

client/config/config.go
```diff
@@ -276,6 +276,16 @@ func (c *Config) FixupAndValidate() error {
 	} else {
 		return err
 	}
+
+	// XXX: Our SOCKS proxy client does not yet support UDP
```
david415, line 280: Can we add a link to a ticket to fix this issue?
// TODO(masala): Make SOCKS proxy client work with UDP. ticket number blah

core/pki/document.go
```diff
@@ -313,19 +313,26 @@ var (
 	// a name server lookup.
 	TransportTCP Transport = "tcp"
 
+	// TransportWS is Websocket
+	TransportWS Transport = "ws"
```
david415, line 317: does this even work? if not, please remove

core/pki/document.go
```diff
@@ -313,19 +313,26 @@ var (
...
+
 	// InternalTransports is the list of transports used for non-client related
 	// communications.
-	InternalTransports = []Transport{TransportTCPv4, TransportTCPv6}
+	InternalTransports = []Transport{TransportTCPv4, TransportTCPv6, TransportQUIC}
 
 	// ClientTransports is the list of transports used by default for client
 	// to provider communication.
-	ClientTransports = []Transport{TransportTCP, TransportTCPv4, TransportTCPv6}
+	ClientTransports = []Transport{TransportTCP, TransportTCPv4, TransportTCPv6, TransportQUIC, TransportWS}
```
david415, line 335: remove TransportWS if not working

http/common/quic.go
```diff
@@ -0,0 +1,196 @@
...
+		// TODO: support QUIC Datagram HTTP Connect proxy dialers
+		_, err = dialFn(ctx, "udp", u.Host)
+		if err != nil {
+			return nil, err
+		}
+		// However the default net.Dialer returns a connected socket that
+		// makes quic.Dial fails with error: use of WriteTo with pre-connected UDP ..
+		// because we need to bind the local listening socket with net.ListenUDP.
+
+		// XXX: verify that quic.Dial does not leak DNS
```
david415, line 181: can we add a ticket link here? and your name
// TODO(masala): verify ... and ticket # blah

david415 (changes requested):

.github/workflows/go.yml
mixmasala: If -onlyTransports is not specified, then mixes listen on all protocols.

When outgoing connections are established (server/internal/outgoing/outgoing_conn.go:164) the Address order depends on the order populated from ranging over the MixDescriptor Addresses (core/pki/descriptor.go:74)

authority/voting/server/server.go
mixmasala: I believe that this code should work although there aren't tests or a websocket client as of yet.
I'll remove this function from this PR and put it in another PR.

authority/voting/server/wire_handler.go
mixmasala: correct

client/config/config.go
mixmasala: Tor doesn't support UDP transport either, so the fail-closed behavior is probably what we want.

.github/workflows/go.yml
mixmasala: it should show if there is a regression/breakage related to a single transport type - because if onlyTransports is not specified all transports are used.

leif:
> `http` doesn't seem like the right name for the scheme, since it is really just QUIC. should it be `quic://` instead?

i see you renamed the transport to `quic`, but `quic.go` is still underneath `http/common`.

mixmasala:
> > `http` doesn't seem like the right name for the scheme, since it is really just QUIC. should it be `quic://` instead?
> 
> i see you renamed the transport to `quic`, but `quic.go` is still underneath `http/common`.

resolved

PR #289, opened 2023-05-21
core: Replace sha3 with blake2b, replace blake2s with blake2b

david415:
Let's migrate away from NIST cryptographic primitives. Replacing SHA3 with Blake2b is good. Let's do good.


ioerror:
Review is partially blocked by `Codecov / codecov/project` failing.

Is there any reason to use blake2b rather than blake3? BLAKE3 is many times faster than blake2b and it has no incompatible variants. 

Looking at https://github.com/BLAKE3-team/BLAKE3/ we see that BLAKE3 is extremely fast. 

There are some  design choices made with BLAKE3 that are discussed by a BLAKE3 coauthor in https://eprint.iacr.org/2019/1492.pdf that might be surprising. The argument for reducing the number of rounds based on an analysis of public cryptanalysis is an interesting one but ultimately it isn't very compelling

In that short paper, which is a good read, a few things are stated that seem surprising. For example page two doesn't really represent the history of DES fairly as it says that "neither DES nor GOST are practically broken by cryptanalysis." This is a very limited understanding of cryptanalysis where the key size being weakened _on purpose by NSA_ from 64 to 56 bits is somehow not a valid Cryptanalysis. It seems obvious that they looked at how they could break it and designed it exactly to be breakable by them, but not others. There are even public records of NSA discussing this exact strategy. The public lore is that NSA helped strengthen DES but that lore usually ignores that NSA didn't need any fancy attacks, the "enhancements" they made bought them decades of cover, and ultimately even at the time NSA people knew that a small enough key size was enough for them to practically break DES when they needed to do so. It was a NOBUS style argument without acknowledging that NOBUS related to key size reductions eventually became everybody. It's a funny notion of NOBUS that probably wasn't even limited to NSA at the time, and is now well within hobby budgets. People like Martin Hellman raised this concern in the _1970s_  ( see e.g.: https://cryptome.org/hellman/hellman-nsa.htm ) and this paper basically ignores that history entirely.

Page nine and page ten of the paper essentially dismiss large-scale adversaries including NSA and Mossad and in a pretty hand-wavy manner. For example, the paper only addresses FLAME  malware in a footnote. FLAME authors executed a practical collision attack (by NSA/CIA and lsraeli military) against a Microsoft signing service because Microsoft continued to use a broken hash function (MD5) long after it should have been retired. What is noteworthy is that the attack used _a new chosen-prefix collision attack_ that hadn't previously been seen in public. The footnote does concede that this is a counter example but the author doesn't actually make a compelling argument that FLAME is an exceptional case or that this won't happen again and again.

A cryptographic weakness will be exploited, even in public, and it likely will be done in a way that shows exceptionally skilled adversaries who bring novel attacks to the table. The argument in the paper essentially says that we should ignore such large-scale adversaries and their actual demonstrated capabilities and go with the public academic analysis. The idea is that if you're targeted, you're doomed, so lets at least save some CPU cycles eh?

This take is a red herring. Microsoft was targeted successfully because MD5 was broken _in a way that wasn't previously known_ to anyone in public, and then it was exploited in a novel manner in public. The value of disclosing a new public attack vector to complete the attack was surprising to many at the time and it shows that NSA had very smart, very talented people working on collision research. This is _how they were targeted_ and _why_ they were doomed. That there are other vectors of exploitation possible isn't a reason to leave _that_ vector on the table. It doesn't follow that reducing the rounds of MD5 or any other hash function, broken or unbroken in public would have been a good idea. The paper doesn't advocate that for MD5 but the general argument suggests that we should take things that aren't broken yet, and assume that we can safely make them _weaker_.

Furthermore, the paper's conclusion that we can safely reduce rounds doesn't really follow unless we assume that we will know or be able to successfully estimate the delta between public and private attacks. I wouldn't take that bet. If the difference is a few extra cycles vs a practically broken system, I'm not sure anyone should do so if their primary concern is _security_.

A desire for efficiency here without knowing the private knowledge delta of large-scale adversaries is dangerous. Large-scale adversaries have billions to spend on _active_ attacks and total impunity for doing so. Microsoft obviously should have switched away from MD5 to something more conservatively designed, and not something weaker with better performance. Page eleven conveniently leaves out MD5 from the history. Page 14 at least makes clear that this proposal wouldn't be unanimously supported though this isn't by acknowledging that the delta revealed by FLAME exploitation is a non-negligible matter. 

It is only a matter of time before other large-scale adversaries have similar capabilities, and in some cases it's plainly obvious that there are peer level adversaries. The original MD5 attacks by Wang et al for example didn't come from NSA and while FLAME's novel exploit did, why intentionally reduce rounds and subsequently weaken anything? It's hard enough to actually secure things, why be up against the edge with an unknown delta of attacker capabilities? Performance, of course!

Performance and a grounding of security on (carefully selected) facts - which is again not incorrect as it assumes that since we don't know private capabilities, the assumption is simply to conclude that the delta is irrelevant! If _those kinds of agencies_ target you, you're doomed after all!

Attempts to settle on a non-zero delta is even disparagingly called "_Cryptographic numerology_"  if someone should even try to guess anything at all. On page nine such _numerology_ is explicitly laid out as *irrational* and *ineffective*: "_We therefore believe that assessing security by comparing numbers without rooting these numbers in reality—what Grigg and Gutmann called “cryptographic numerology” [35]—is not a
rational nor effective approach to risk._"  

It seems rational to assume the delta of knowledge isn't zero based on _public evidence_ including _budgets_ and _known incidents_ among other matters. Thus the authors might look at history to tell us how far ahead some large-scale attackers are in their capabilities; they do not do so. The resulting number is not zero, and no such analysis is presented in the paper.

It is also an effective approach to mitigate risk by favoring _enhancement of security_ at the cost of performance, especially if the performance numbers likely already saturate a system's other resources. For example, if you can't do 15Gb/s of hashes in software for a system that has a 10Gb/s network connection, is that a problem? Most of the computations won't be spent on hashing but even if they were, that isn't a relevant bottleneck, and then it is possible that _a security vulnerability has been introduced_ to optimize for something that is largely irrelevant.

Despite the above arguments from the referenced paper that seem counter to many of the intuitions of some Katzenpost folks, BLAKE3 is probably a reasonable choice for at least two reasons. The first reason is that the _software_ performance and the overall design is impressive. It is a versatile design giving all kinds of useful definitions for various things that Katzenpost needs. Secondly, if this is an incorrect choice, hopefully any exploitation can be detected and then it can serve as an interesting example to others just as FLAME did for MD5. Katzenpost should aim to be positioned to detect such an attack and have a policy of disclosing it before that happens, if it happens.

See https://github.com/BLAKE3-team/BLAKE3 and additionally pure golang implementations of blake3 are available ( https://github.com/lukechampine/blake3 and https://github.com/zeebo/blake3 )

Related is the issue https://github.com/golang/go/issues/36632 where there is a discussion to bring blake3 into the standard library for golang.

threebithacker:
Operationally, depending on something in upstream `crypto/` is nicer because it's less likely to randomly make breaking API changes, that is one argument in favor of blake2b for now. Blake3 is relatively new and it sounds like there are concerns; I don't think we have any evidence of the hashing here being a significant bottleneck (warranting rolling the dice on something new).

If we ever get such numbers we could replace *all* of our hashing with blake3, but for now the exercise was just getting rid of `sha3`(using blake2b) and `shake`(using blake2x) because the sha3 dep stinks.

We probably should get rid of our use of blake2s in favor of blake2b at some point too.

threebithacker:
PR looks good to me but I suggest we hold off for a few days and see if we need to make any more updates to `main` to resolve the current operational issues. Then we can make this breaking change once it's settled down.

mixmasala (changes requested):
this PR replaces sha3 in core/pki/sharedrandom.go but in authority/voting/server/state.go sha3 is still used for part of the shared random computation. Shouldn't we remove the sha3 dependency there as well?

mixmasala (approved):
this PR includes a few extra features such as removing the nonvoting authority, which I'm not particularly opposed to, but is somewhat offtopic.

mixmasala:
Merging this as-is but I think we need to update the shared random computation documentation to mention the hash function change. This also removes the nonvoting authority but since no-one wants to maintain that, so be it.

mixmasala:
I'm sorry, the shared random docs were updated

PR #292, opened 2023-05-21
Rm nonvoting

david415:


ioerror:
I wanted to review this but it appears that the CI didn't pass.

threebithacker:
CI complains about things like
```
Error: ../client/config/config.go:28:2: no required module provides package github.com/katzenpost/katzenpost/authority/nonvoting/client; to add it:
```

This needs a `git grep` pass:
```shell
git grep -i non-voting
git grep -i nonvoting
```

leif:
are we really sure we want to remove the nonvoting authority? for quickly testing anything unrelated to consensus is was pretty handy when it worked. i think we only gave up on keeping it working toward the very end of the big sprint in december so it is probably not *too* much work to make it usable again.

david415:
> are we really sure we want to remove the nonvoting authority? for quickly testing anything unrelated to consensus is was pretty handy when it worked. i think we only gave up on keeping it working toward the very end of the big sprint in december so it is probably not _too_ much work to make it usable again.

It absolutely does not make sense from a code maintenance point of view. However I am probably the only one to experience that particular source of pain because of my many changes to our core cryptography.

mixmasala:
closing this PR since nonvoting got removed in #289
If we want to bring it back, open a PR.

PR #295, opened 2023-05-24
Add -v version info flag to main funcs

david415:


ioerror (approved):
LGTM if adding a dependency isn't an issue; without adding a dependency, you could add an import of `runtime/debug`  and then call `debug.ReadBuildInfo()` to return the same data.

Issue #300, opened 2023-05-31
server: cborplugin commandIO only supports cborplugin.Request and cborplugin.Response Command types


mixmasala:
Although the server plugin can implement cborplugin.ServicePlugin and return a cborplugin.Command, the plugin only works if it returns a cborplugin.Response.

This is suboptimal because it means responses need to be marshal'ed twice.
Instead, the server worker should return the serialized bytes.

PR #301, opened 2023-05-31
server/cborplugin: Support returning raw plugin responses

mixmasala:
https://github.com/katzenpost/katzenpost/issues/300

This avoids deserializing a cborplugin.Response (which just wraps Payload []byte) and then its enclosed field. Client plugins should know what types of response to expect from the services they query.

I'm not sure yet what else this change will break downstream: I think it'll require changing a few kaetzchen clients so that they deserialize responses directly into the appropriate types rather than deserializing into a cborplugin.Response and then deserializing the Response.Payload into whatever was returned.


server/internal/provider/kaetzchen/cbor_plugins.go
```diff
@@ -136,34 +136,33 @@ func (k *CBORPluginWorker) processKaetzchen(pkt *packet.Packet, pluginClient *cb
...
+	}
+	if len(payload) > k.geo.UserForwardPayloadLength {
+		// response is probably invalid, so drop it
+		k.log.Errorf("Got response too long: %d > max (%d)",
+		len(payload), k.geo.UserForwardPayloadLength)
+		instrument.KaetzchenRequestsDropped(1)
+		return
+	}
+	// Iff there is a SURB, generate a SURB-Reply and schedule.
+	if surb != nil {
```
threebithacker, line 153: if we just had
```go
if surb == nil {
	k.log.Debugf("No SURB provided: %v", pkt.ID)
	return
}
```
we could save some indentation below.

server/internal/provider/kaetzchen/cbor_plugins.go
```diff
@@ -136,34 +136,33 @@ func (k *CBORPluginWorker) processKaetzchen(pkt *packet.Packet, pluginClient *cb
...
+	if err != nil {
+		k.log.Errorf("Failed to Marshal CBOR Plugin response: %v", err)
+		instrument.KaetzchenRequestsDropped(1)
+		return
+	}
+	if len(payload) > k.geo.UserForwardPayloadLength {
+		// response is probably invalid, so drop it
+		k.log.Errorf("Got response too long: %d > max (%d)",
+		len(payload), k.geo.UserForwardPayloadLength)
+		instrument.KaetzchenRequestsDropped(1)
```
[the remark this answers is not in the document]
threebithacker: Since we now have a way to figure out what the max permitted length is (`k.geo.UserForwardPayloadLength`), would it makes sense to add that to the protocol between the provider and the backend service? Like a field passing on `UserForwardPayloadLength: k.geo.UserForwardPayloadLength` so the service has a chance to send a reply to the client saying "sorry I can't give you a response because it would be too long" or whatever, without having to read the KP config?

server/internal/provider/kaetzchen/cbor_plugins.go
```diff
@@ -136,34 +136,33 @@ func (k *CBORPluginWorker) processKaetzchen(pkt *packet.Packet, pluginClient *cb
...
 	}
 	cborResponse := <-pluginClient.ReadChan()
-	switch r := cborResponse.(type) {
-	case *cborplugin.Response:
-		if len(r.Payload) > k.geo.UserForwardPayloadLength {
-			// response is probably invalid, so drop it
-			k.log.Errorf("Got response too long: %d > max (%d)",
-				len(r.Payload), k.geo.UserForwardPayloadLength)
-			instrument.KaetzchenRequestsDropped(1)
+	payload, err = cborResponse.Marshal()
```
threebithacker, line 139: Reusing `payload` here is slightly confusing. It would be less error-prone if we used separate variables for the request payload and the response payload.

threebithacker (changes requested):
Minor points, otherwise looks good to me.

server/internal/provider/kaetzchen/cbor_plugins.go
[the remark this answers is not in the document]
mixmasala: Probably the ServerPlugin method  RegisterConsumer() should be used by client plugins to learn the maximum response sizes. #368

server/internal/provider/kaetzchen/cbor_plugins.go
mixmasala: this got refactored elsewhere and merged from main.

Issue #302, opened 2023-06-01
dirauth crashes


david415:
```

16:50:04.147 DEBU state: authority: Wakeup due to voting schedule.  
16:50:04.147 DEBU state: Current epoch 157076, remaining time:  
9m55.85207652s  
panic: runtime error: invalid memory address or nil pointer dereference  
\[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x68556a\]  
  
goroutine 14 \[running\]:  
[github.com/katzenpost/katzenpost/authority/voting/server.(\*state).getDocument(0xc000002140](http://github.com/katzenpost/katzenpost/authority/voting/server.(*state).getDocument(0xc000002140),  
{0x0, 0x0, 0xc0019936e0?}, 0x0, {0xc003fbd4e0, 0x20, 0x20})  
  
/home/katzenpost/katzenpost/authority/voting/server/state.go:516 +0x38a  
[github.com/katzenpost/katzenpost/authority/voting/server.(\*state).getMyConsensus(0xc000002140](http://github.com/katzenpost/katzenpost/authority/voting/server.(*state).getMyConsensus(0xc000002140),  
0x26595)  
  
/home/katzenpost/katzenpost/authority/voting/server/state.go:418 +0x3fe  
[github.com/katzenpost/katzenpost/authority/voting/server.(\*state).fsm(0xc000002140)](http://github.com/katzenpost/katzenpost/authority/voting/server.(*state).fsm(0xc000002140))  
  
/home/katzenpost/katzenpost/authority/voting/server/state.go:221 +0x19b  
[github.com/katzenpost/katzenpost/authority/voting/server.(\*state).worker(0xc000002140)](http://github.com/katzenpost/katzenpost/authority/voting/server.(*state).worker(0xc000002140))  
  
/home/katzenpost/katzenpost/authority/voting/server/state.go:149 +0x9f  
[github.com/katzenpost/katzenpost/core/worker.(\*Worker).Go.func1()](http://github.com/katzenpost/katzenpost/core/worker.(*Worker).Go.func1())  
         /home/katzenpost/katzenpost/core/worker/worker.go:39 +0x5a  
created by [github.com/katzenpost/katzenpost/core/worker.(\*Worker).Go](http://github.com/katzenpost/katzenpost/core/worker.(*Worker).Go)  
         /home/katzenpost/katzenpost/core/worker/worker.go:37 +0xb9
```

threebithacker:
[...]

PR #306, opened 2023-06-01
tallyVotes: return err when decoding bs (params) fails

threebithacker:
see #302 and #305 
We weren't returning the `err` so we got the nondescript catch-all error in this case.


david415 (approved):

threebithacker:
Codecov CI complained about less than 47% test coverage. I overrode that since the new behaviour seemed superior to just silently throwing away the error.

PR #316, opened 2023-06-08
authority: add self to list of authorities

mixmasala:
when an authority bootstraps it uses the voting client to see if the network has a consensus already, but does not consider its own signature.


authority/voting/server/state.go
```diff
@@ -1831,10 +1831,12 @@ func (s *state) backgroundFetchConsensus(epoch uint64) {
 	_, ok := s.documents[epoch]
 	if !ok {
 		go func() {
+			self := &config.Authority{Identifier: s.s.cfg.Server.Identifier, IdentityPublicKey: s.s.IdentityKey(), LinkPublicKey: s.s.linkKey.PublicKey(), Addresses: s.s.cfg.Server.Addresses}
+			allAuthorities := append(s.s.cfg.Authorities, self)
```
threebithacker, line 1835: what happens if it's already *in* the config, then we end up with two copies of ourself?
It seems like we should first check that we aren't in the config. Possibly we should just panic if the config is wrong?
Ie if we think we are an authority, but nobody else has a config that thinks that, trying to participate seems pointless.

threebithacker (dismissed):

authority/voting/server/state.go
mixmasala: from client.New():
```
// New constructs a new pki.Client instance.
func New(cfg *Config) (pki.Client, error) {
        if cfg == nil {
                return nil, fmt.Errorf("voting/Client: cfg is mandatory")
        }
        if err := cfg.validate(); err != nil {
                return nil, err
        }

        c := new(Client)
        c.cfg = cfg 
        c.log = cfg.LogBackend.GetLogger("pki/voting/Client")
        c.pool = newConnector(cfg)
        c.verifiers = make([]cert.Verifier, len(c.cfg.Authorities))
        for i, auth := range c.cfg.Authorities {
                c.verifiers[i] = auth.IdentityPublicKey
        }
        c.threshold = len(c.verifiers)/2 + 1
        return c, nil
}
```

mixmasala: if the authority were in the configuration twice, it would end up with the # of verifiers and threshold being wrong

threebithacker: Right. But if we didn't magically add ourselves to the config, and instead mandated own pk being in the config, we could use the same config entry for all the auths. That seems a lot easier to manage/troubleshoot.

mixmasala: so, duplicate the key material in the config from the Server section to the Authorities section?

mixmasala:
Ok, I changed the behavior to have the Authorities section of the authority configuration require containing itself, and genconfig to configure the authority in this way.

authority/voting/server/config/config.go
```diff
@@ -472,13 +473,36 @@ func (cfg *Config) FixupAndValidate() error {
 		pkMap[tmp] = v
 	}
 
+	// if our own identity is not in cfg.Authorities return error
+	selfInAuthorities := false
+
+	ourPubKeyFile := filepath.Join(cfg.Server.DataDir, "identity.public.pem")
+	f, err := os.Open(ourPubKeyFile)
```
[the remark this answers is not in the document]
mixmasala: #366

PR #327, opened 2023-06-17
docker/Makefile: add clean-bin target

leif:
this is useful for creating a testnet with an old version and then upgrading it without deleting the network state (as `clean-local` does).

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -10,6 +10,7 @@ help:
 	@echo " status             - show testnet consensus status"
 	@echo " show-latest-vote   - does what it says"
 	@echo " run-ping           - send a ping over the testnet"
+	@echo " clean-bin          - stop, and delete compiled binaries"
 	@echo " clean-local        - stop, and delete data and binaries"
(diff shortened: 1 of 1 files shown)


docker/Makefile
```diff
@@ -197,7 +198,10 @@ clean-image-%:
 	-$(docker) rmi $(patsubst clean-image-%,katzenpost-%,$@)
 	-rm -fv $(patsubst clean-image-%,%,$@).stamp
 
-clean-local: stop
+clean-bin: stop
+	rm -vf $(net_name)/*.$(distro)
```
threebithacker, line 202:
```suggestion
	rm -vf ./$(net_name)/*.$(distro)
```
how about this, then at works we nuke `.//.`? :-)

threebithacker (changes requested):

docker/Makefile
threebithacker: (which should be fine since it's not `rm -r`)

mixmasala (approved):
thanks

PR #330, opened 2023-06-18
Fix 323 handle cborplugin failure

mixmasala:
This PR handles a kaetzchen plugin crash gracefully by unregistering it from the set of services.

### DIFF
--- a/server/cborplugin/client.go
+++ b/server/cborplugin/client.go
@@ -218,6 +218,7 @@ func (c *Client) launch(command string, args []string) error {
 	}
 
 	// proxy stderr to our debug log
+	// also calls Halt() when stderr closes, if the program crashes or is killed
 	c.Go(func() {
 		c.logPluginStderr(stderr)
 	})
--- a/server/cborplugin/socket.go
+++ b/server/cborplugin/socket.go
@@ -122,7 +122,8 @@ func (c *CommandIO) reader() {
 		cmd := c.commandBuilder.Build()
 		err := dec.Decode(cmd)
 		if err != nil {
-			panic(err) // XXX
+			c.Halt()
+			return
 		}
 		select {
 		case <-c.HaltCh():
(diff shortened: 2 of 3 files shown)


mixmasala:
I tested this by starting a dockerized mixnet and then killing one of the cbor plugin workers and verifying the provider node has not crashed.

```
06:27:28.286 DEBU scheduler: Batch processing 1 packets.
06:27:28.286 DEBU scheduler: Enqueueing packet: 1966 delta-t: 10ms
06:27:28.297 DEBU outgoing:1: Sent packet: 1966
06:27:28.324 DEBU outgoing:1: Sent packet: 1958
>>> 06:27:28.366 ERRO client: CBOR plugin worker, command exec error: signal: killed <<<
06:27:28.374 DEBU outgoing:1: Sent packet: 1962
06:27:28.536 DEBU incoming:1: Handing off packet: 1967
06:27:28.536 DEBU crypto:0: Packet: 1967 (Unwrap queue delay: 85.867µs)
06:27:28.536 DEBU crypto:0: Attempting to unwrap packet: 1967
06:27:28.537 DEBU crypto:0: Packet: 1967 (Unwrap took: 931.542µs)
06:27:28.537 DEBU crypto:0: Packet: 1967 (IsReplay took: 155.931µs)
```

However, in the next PKI, the descriptor still advertised the crashed service...

mixmasala:
Oops. I called Wait() from inside a routine started by worker.Go().
I refactored this into a method unregister()

mixmasala:
I've tested that this works by killing the service plugins and verifying that the service is no longer published in the PKI

threebithacker (approved):
Code looks fine to me, but if you think that the invariants you manually tested are important to preserve, perhaps we should add a CI test for that?

mixmasala:
"Draft status because we don't know if this starts panic()ing all over"
This looks OK to merge to me. It shouldn't do anything if the code is correct, right  :)?

PR #332, opened 2023-06-18
authority/voting/client: #287 make pki retry also verify document

### LINKED ISSUE #287: authority/client does not retry on GetConsensus threshold failure
```13:31:04.446 NOTI katzenpost/client: 😼 Katzenpost is still pre-alpha.  DO NOT DEPEND ON IT FOR STRONG SECURITY OR ANONYMITY. 😼
13:31:06.216 NOTI pki/voting/client/connector: got response from annares to GetConsensus(156928) (attempt 0, err=<nil>, res=Ok)
13:31:06.225 ERRO pki/voting/Client: VerifyThreshold failure: 0 good signatures, 6 bad signatures: threshold failure
panic: failed to create session: voting/Client: Get() invalid consensus document: threshold failure

goroutine 1 [running]:
main.main()
	/build/katzenpost/ping/main.go:75 +0x433
```

mixmasala:
the connector doesn't have a reference to verifiers, so move the loop outside of fetchConsensus so that a retry can happen if VerifyThreshold fails. For example, if a buggy/broken authority serves an invalid document.

fixes #287


threebithacker (changes requested):

authority/voting/client/client.go
```diff
@@ -326,61 +321,65 @@ func (c *Client) Get(ctx context.Context, epoch uint64) (*pki.Document, []byte,
...
 	// Dispatch the get_consensus command.
-	resp, err := c.pool.fetchConsensus(ctx, linkKey, epoch)
-	if err != nil {
-		return nil, nil, err
-	}
+	// Try to fetch a consensus up to threshold number of times
+	for i := 0; i < len(c.cfg.Authorities)/2; i++ {
+		resp, err := c.pool.fetchConsensus(ctx, linkKey, epoch)
+		if err != nil {
+			return nil, nil, err
```
threebithacker, line 328: is this retried somewhere else, or should we just log + `continue` here if connecting to the auth failed?

authority/voting/client/client.go
```diff
@@ -326,61 +321,65 @@ func (c *Client) Get(ctx context.Context, epoch uint64) (*pki.Document, []byte,
 	defer close(doneCh)
 
 	// Dispatch the get_consensus command.
-	resp, err := c.pool.fetchConsensus(ctx, linkKey, epoch)
-	if err != nil {
-		return nil, nil, err
-	}
+	// Try to fetch a consensus up to threshold number of times
```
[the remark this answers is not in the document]
mixmasala: The simplest thing to do is just to try all the authorities.

authority/voting/client/client.go
mixmasala: oops. I've refactored this again, see commit below.

david415 (approved):
looks good to me

PR #335, opened 2023-06-19
Fix transient errors in TestBadCertificate:

threebithacker:
```
 --- PASS: TestWrongCertificate (0.92s)
=== NAME  TestBadCertificate
    cert_test.go:87:
         Error Trace:
/home/runner/work/katzenpost/katzenpost/core/crypto/cert/cert_test.go:87
         Error:       An error is expected but got nil.
         Test:        TestBadCertificate
--- FAIL: TestBadCertificate (0.61s)
```

This happens when `certificate[1000] == 235` in which case no modification is made, and we unexpectedly get an OK from `Verify()`. This used to happen in 1/256 test executions.

Also bumps the expiry by one so we're not impacted by epoch rollover between `Sign()` and `Verify()`, which we haven't run into, but could happen one day.

### DIFF
--- a/core/crypto/cert/cert_test.go
+++ b/core/crypto/cert/cert_test.go
@@ -75,13 +75,18 @@ func TestBadCertificate(t *testing.T) {
 	signingPrivKey, signingPubKey := Scheme.NewKeypair()
 
 	current, _, _ := epochtime.Now()
+	validBeforeEpoch := current + 2
+	// +2 so we it's not impacted by epoch rollover in between Sign()
+	//  and Verify(): it's valid during [current, current+1]
 
 	certified := []byte("hello, i am a message")
 
-	certificate, err := Sign(signingPrivKey, signingPubKey, certified, current+1)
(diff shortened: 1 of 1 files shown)


core/crypto/cert/cert_test.go
```diff
@@ -75,13 +75,18 @@ func TestBadCertificate(t *testing.T) {
...
 	certified := []byte("hello, i am a message")
 
-	certificate, err := Sign(signingPrivKey, signingPubKey, certified, current+1)
+	certificate, err := Sign(signingPrivKey, signingPubKey, certified, validBeforeEpoch)
 	require.NoError(t, err)
 
-	certificate[1000] = 235 // modify the signed data so that the Verify will fail
+	// modify the signed data so that the Verify will fail.
+	// XOR ensures modification:
+	certificate[1000] ^= 235
```
mixmasala, line 89: amazing.

mixmasala (approved):
great work.

PR #340, opened 2023-06-24
github-ci: run package unit tests in separate steps

mixmasala:
split unit tests into separate github-ci steps so that failures are easier to spot


mixmasala:
this is unfortunate, but I don't see another way to not have to wade through all the concatenated test logs.

PR #341, opened 2023-06-26
Fix test benchmark panic

mixmasala:
This should resolve the benchmark test failure where the client was only waiting for the PKI document fetch rather than provider connection before starting the benchmark, which caused a race.


mixmasala:
#342

PR #348, opened 2023-06-27
docs/specs/pki: update pki spec of hash function for shared random

mixmasala:


mixmasala:
Is there any other documentation that should be updated?

PR #390, opened 2023-10-02
By default record prometheus metrics

david415:


mixmasala (commented):
why do you remove the ability to build without prometheus? can we instead add:
//go:build !noprometheus
to prometheus.go
and
//go:build noprometheus
to prometheus_dummy.go

see:
https://github.com/katzenpost/katzenpost/tree/build_server_with_prometheus_by_default

I looked over the set of exposed counters - can we save / snapshot metrics somewhere? afaik this will enable performance counters but does not save the metrics in a time series format. How does a server operator save / view these metrics? Should we write metrics to the server logfile periodically?
// Init initialize instrumentation
func Init() {
        // Register metrics
        prometheus.MustRegister(deadlineBlownPacketsDropped)
        prometheus.MustRegister(incomingConns)
        prometheus.MustRegister(invalidPacketsDropped)
        prometheus.MustRegister(outgoingConns)
        prometheus.MustRegister(ingressQueueSize)
        prometheus.MustRegister(outgoingPacketsDropped)
        prometheus.MustRegister(packetsDropped)
        prometheus.MustRegister(packetsReplayed)
        prometheus.MustRegister(ignoredPKIDocs)
        prometheus.MustRegister(kaetzchenRequests)
        prometheus.MustRegister(kaetzchenPacketsDropped)
        prometheus.MustRegister(kaetzchenRequestsDropped)
        prometheus.MustRegister(kaetzchenRequestsDuration)
        prometheus.MustRegister(kaetzchenRequestsFailed)
        prometheus.MustRegister(mixPacketsDropped)
        prometheus.MustRegister(mixQueueSize)
        prometheus.MustRegister(pkiDocs)
        prometheus.MustRegister(cancelledOutgoingConns)
        prometheus.MustRegister(fetchedPKIDocs)
        prometheus.MustRegister(fetchedPKIDocsDuration)
        prometheus.MustRegister(failedFetchPKIDocs)
        prometheus.MustRegister(failedPKICacheGeneration)
        prometheus.MustRegister(invalidPKICache)

        // Expose registered metrics via HTTP
        http.Handle("/metrics", promhttp.Handler())
        go http.ListenAndServe("127.0.0.1:6543", nil)
}

david415:
> why do you remove the ability to build without prometheus? can we instead add: //go:build !noprometheus to 

i remove the ability to build without prometheus because there's no reason not to.

mixmasala (commented):
do we want to keep the timers? I think yes, because it helps us debug slow handlers. Those data races were fixed in: 372918b95a6f1a235d3b2976025ff67e654459f5 but revert in 5d18820fa8a622b7f6148ac2f8b897956590b152 lost those fixes, basically a timer needs to be instantiated within scope of processKaetzchen. I've opened https://github.com/katzenpost/katzenpost/issues/398 to track this.

mixmasala:
I'm trying to figure out if the prometheus collector supports querying unix sockets, but I haven't found any examples of this yet.

leif:
> I'm trying to figure out if the prometheus collector supports querying unix sockets, but I haven't found any examples of this yet.

it isn't possible in prometheus directly (yet) so we'd need a proxy shim like socat: https://github.com/prometheus/prometheus/issues/12024

Dmitry1987:
this breaks deployments from manifest, for all who don't use the helm chart. why prometheus socket is mandatory? imho should be optional 🤔 
```
F1019 14:02:56.444660       7 main.go:135] Error creating prometheus collector:  listen unix /tmp/nginx/prometheus-nginx.socket: bind: no such file or directory
```

Dmitry1987:
lol what repo is this, how did I end up here from the nginx ingress repo 😆  my bad.. i thought it's some change in the ingress code.

PR #393, opened 2023-10-03
bench/bench_test: use SphinxGeometry from client configuration

mixmasala:
this fixes bench test, if the parametized network uses a different geometry


david415:
thanks for fixing this!

PR #401, opened 2023-10-05
docker: add prometheus instance to deployment

mixmasala:
this adds an instance of prometheus and generates a configuration for it so that each mix node can be polled.
TODO: figure out what sampling interval to use, and how to best integrate into CI


ioerror (approved):
LGTM

PR #402, opened 2023-10-05
docker: fix DOCKER_HOST env as newer versions of docker-compose needs a URL

mixmasala:
I noticed that docker-compose stopped working when upgrading from alpine 3.17 to alpine 3.18. (docker-cli-compose 2.15.1-r3 -> 2.17.3-r6)


mixmasala:
https://github.com/docker/compose/issues/8733

PR #421, opened 2023-10-15
omit hasEnoughDescriptors

mixmasala:
Authorities can learn about MixDescriptors from other Authorities


mixmasala:
I noticed that auths set their sleep time before signing documents, so I fixed that too

PR #434, opened 2023-10-18
readme: add more details about the pki signature scheme

david415:
fixes https://github.com/katzenpost/katzenpost/issues/433


README.md
```diff
@@ -183,10 +183,28 @@ QROM.
...
+
+Note that we could have just used Sphincs+ since hash based signature schemes have a
+solid theoretical foundation, however there can still be mistakes in the hyper tree design
+and so pairing it with ed25519 is prudent defense in dept.
+
+Clients also use this signature scheme to verify PKI documents.
+
+Currently, we are using the *SHAKE-256f* parameterization of Sphincs+ however we will
+soon migrate away from that parameter set as it contains a NIST primitive we would
+like to avoid.
```
mixmasala, line 206: We use some terminology here that isn't yet introduced, e.g. terms PKI and "mix descriptor" should be explained:
e.g.
"Clients and mix nodes can talk to ..." 
with
"Mix nodes publish their IP addresses and per-epoch keys in a self-signed "Mix Descriptor" to the directory authority nodes, which vote and produce a multisignature PKI document each Epoch. Clients receive this document each epoch and verify there is a threshold of signatures from the authorities.

mixmasala: Is there a particular reason we want to avoid SHAKE-256f?

david415: > Is there a particular reason we want to avoid SHAKE-256f?

usually cypherpunks don't need to explain to other cypherpunks about avoiding NIST cryptographic primitives, yet that's what i'm doing here. we should not want to go to bed with NIST.

Issue #432, opened 2023-10-17
change sphincs+ parameters to not use NIST primitives


david415:
Let's not use SHAKE in our Sphincs+. there are many other parameters to choose from. so far Sphincs+ is parameterized with sha2, shake and haraka. Let's use Haraka for the win!

david415:
We decided that our SHAKE params for Sphincs+ are the most conservative. If we want higher performance we should select a SHA2 param set.

PR #435, opened 2023-10-18
Fix sphincs plus params.2

david415:
Fix Sphincs+ params to not use NIST hash! fixes https://github.com/katzenpost/katzenpost/issues/432

Since this breaks compatibility... consider merging this later in the week or later in the month right before a coordinated upgrade.


mixmasala (commented):
Can we verify that build outputs are using hardware aes?
per the spec https://sphincs.org/data/sphincs+-specification.pdf 
"We used the same system as described above, this time including the
march=native" 

"SPHINCS+ implementations are
thus free of the two most notorious sources of timing variation. An exception is potentially
SPHINCS+-Haraka, because Haraka is based on AES, which is well known to exhibit timing
vulnerabilities in software implementations [3, 15, 5, 14]. Clearly, SPHINCS+-Haraka should
only be used in environments that support AES in hardware"

PR #437, opened 2023-10-18
Add bench tests for signature scheme

david415:


mixmasala (changes requested):
can has copyright header too?

mixmasala (approved):

PR #438, opened 2023-10-20
core/crypto: fixup KEM combiner

david415:
works. merge me


david415:
NOTE that the TestHybridKEMVectors test is failing. This makes sense since I just changed the KEM combiner. However we don't want to fix the test vectors until we also fix the NIKE to KEM adapter.

ioerror (approved):

PR #485, opened 2024-01-09
cborplugin/socket: retry socket dial failures in case of race on plugin Accept

### LINKED ISSUE #477: mixserver cborplugin races socket creation
when a cborplugin.ServerPlugin is launched, it must emit the socketfile to stdout, and wait for a connection to this socket.
If the mixserver tries to connect to the unix socket before the plugin is ready to Accept a connection, the mix server fails to start.

The behavior of several plugins is to emit the socket to stdout and then Accept the connection, which creates a race between the mix server and the plugin.
Because Accept is a blocking call, the plugin must instead Accept from another routine before printing the socketfile to stdout. However, as Accept is a blocking call, there is not an obvious way to coordinate between the routines.

mixmasala:
fixes #477, as long as plugins Accept within 3 seconds of writing their socketfile to stdout


mixmasala:
This is described in #477, but tdlr:

1. the plugin is launched by the mix server, and it chooses a path for a unix socket, and passes this to a cborplugin.CommandIO instance via its Start method
2. the plugin writes the path to stdout and calls the Accept method of the CommandIO instance
3. the mix server dials the unix socket

The problem is that the call to Accept blocks, so starting the listener in another routine isn't guarranteed to be ready before the socketfile is connected by the mixserver.
Occasionally this fails to happen in the right order and the mix server will panic on startup. This happens more frequently if the plugin handler is slow to startup.

The workaround/fix implemented here is for the mix server to retry dialing the socket a few times.

PR #487, opened 2024-01-10
sphinx: improve tests

david415:
This change makes sure that the Sphinx unit tests are testing for the correct packet size at each stop in the packet unwarp. Here we also add a SURB test for KEM Sphinx. And we fix the CTIDH Sphinx tests.


ioerror (approved):
LGTM

PR #489, opened 2024-01-11
Use latest highctidh

david415:
if all tests pass then please merge me


ioerror (approved):
Ideally the next improvement should be to test and use each highctidh field size.

PR #502, opened 2024-01-23
Add ability to start and stop Kaetzchen and CBORPluginKaetzchen to Provider

mixmasala:
Adds THWACK commands to start and stop mix server plugins. This is useful for developing plugins, as it enables the devloper to start and stop plugins without waiting for a test network to bootstrap.


docs/handbook/mix_server.md
```diff
@@ -433,3 +433,15 @@ SEND_RATE 30
 ```
 SEND_BURST 4
 ```
+
+- `START_KAETZCHEN` - Start the Kaetzchen identified by capability:
```
david415, line 437: can we make the command easier to type and shorter? maybe like just "START"?

mixmasala: STARTK? I'm not sure START is unambiguous enough.

david415: is there already another start? i think it's pretty unambiguous if we document it.

mixmasala: Not yet, but we may want a graceful STOP command so that mixes can shutdown at end-of-epoch (and not publish another descriptor - so in that case STOP might be used for that purpose)

david415 (approved):

david415 (approved):

david415:
if the tests pass, merge asap

PR #503, opened 2024-01-24
server/internal/mixkey: eliminate mixkey perisistence

mixmasala:
This eliminates problematic MixKey persiistence to disk. Previously, the mix server would write key material to disk, and not erase it until the mix service was restarted.
Instead, mix servers should do graceful upgrades by halting the mix at end of epoch and restarting into a new epoch with new in-memory key material.

### DIFF
--- a/server/internal/mixkey/mixkey.go
+++ b/server/internal/mixkey/mixkey.go
@@ -110,11 +110,9 @@ func (k *MixKey) IsReplay(rawTag []byte) bool {
 	defer k.Unlock()
 
 	// If the filter is saturated then probability of a false replay is increased
-	// XXX: the filter size should be turned for the maximum line rate expected so that this does not happen
+	// XXX: the filter size should be tuned for the maximum line rate expected so that this does not happen
 	if k.f.Entries() >= k.f.MaxEntries() {
-		// XXX: does not have a logger...
-		panic("bloom filter size too small")
-		//k.log.Errorf("bloom filter reached satruation: '%v'", k.f.MaxEntries())
+		panic("MixKey bloom filter size too small")
 	}
 	if !k.f.TestAndSet(tag[:]) {
 		return false
--- a/server/internal/mixkey/mixkey.go
+++ b/server/internal/mixkey/mixkey.go
@@ -19,46 +19,20 @@ package mixkey
 
 import (
 	"crypto/sha512"
-	"encoding/binary"
-	"fmt"
-	"math"
-	"os"
-	"path/filepath"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


server/internal/mixkey/mixkey.go
```diff
@@ -136,138 +106,20 @@ func (k *MixKey) IsReplay(rawTag []byte) bool {
...
-func (k *MixKey) testAndSetTagMemory(tag *[TagLength]byte) (bool, bool) {
 	k.Lock()
 	defer k.Unlock()
 
-	// TODO/perf: Perhaps the lock should only cover the bloom filter,
-	// and a sync.Map used for the pending write-back entries.
-
-	// If the filter is saturated then force a database lookup.
+	// If the filter is saturated then probability of a false replay is increased
+	// XXX: the filter size should be turned for the maximum line rate expected so that this does not happen
```
david415, line 113: s/turned/tuned/ spell check

server/internal/mixkey/mixkey.go
```diff
@@ -136,138 +106,20 @@ func (k *MixKey) IsReplay(rawTag []byte) bool {
...
 	if k.f.Entries() >= k.f.MaxEntries() {
-		return true, k.writeBack[*tag]
+		// XXX: does not have a logger...
```
david415, line 115: 1. remove the comment about not having a logger or 2. add the logger. pick one. you must choose between those two options.

server/internal/mixkey/mixkey.go
```diff
@@ -136,138 +106,20 @@ func (k *MixKey) IsReplay(rawTag []byte) bool {
...
+		panic("bloom filter size too small")
+		//k.log.Errorf("bloom filter reached satruation: '%v'", k.f.MaxEntries())
```
david415, line 117: remove the commented out line of code or remove the panic and uncomment the log line. pick one of those two options.

david415 (changes requested):
fix the two little nitpicks and i'll approve and merge this

server/internal/mixkey/mixkey.go
mixmasala: thanks

server/internal/mixkey/mixkey.go
mixmasala: I think panic is more likely to be noticed and acted upon, so I went with that choice.

server/internal/mixkey/mixkey.go
david415: the filter size should be tuned for.... not "turned"

david415 (approved):

Issue #123, opened 2022-12-10
At notice log level, we may log IP addresses


ioerror:
While the consensus is in a failed state, using the ping client to test the network, it generated this log on annares:
```
Dec 10 18:04:49 power9 pq-katzenpost-authority[1126121]: 18:04:49.330 ERRO state: No document for current epoch 145350 generated and never will be
Dec 10 18:04:49 power9 pq-katzenpost-authority[1126121]: 18:04:49.330 ERRO authority: Peer 1.2.3.4:36250: Failed to retrieve document for epoch '145350': authority: Requested epoch will never get a Document
```

I have replaced my IP with 1.2.3.4 in the log above, we should either scrub client IP addresses or simply not log them at all.

PR #504, opened 2024-01-24
authority/voting/server: omit logging PII upon GetConsensus failure

mixmasala:
If there is no document available for the requested epoch, the requesting client IP is being logged.

### DIFF
--- a/authority/voting/server/wire_handler.go
+++ b/authority/voting/server/wire_handler.go
@@ -149,7 +149,6 @@ func (s *Server) onGetConsensus(rAddr net.Addr, cmd *commands.GetConsensus) comm
 	resp := &commands.Consensus{}
 	doc, err := s.state.documentForEpoch(cmd.Epoch)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mixmasala:
fixes #123

authority/voting/server/wire_handler.go
```diff
@@ -149,7 +149,6 @@ func (s *Server) onGetConsensus(rAddr net.Addr, cmd *commands.GetConsensus) comm
 	resp := &commands.Consensus{}
 	doc, err := s.state.documentForEpoch(cmd.Epoch)
 	if err != nil {
-		s.log.Errorf("Peer %v: Failed to retrieve document for epoch '%v': %v", rAddr, cmd.Epoch, err)
```
david415, line 152: Are you sure you want to completely remove the entire log line instead of just modifying it to omit the ip address?

mixmasala: yes

david415 (approved):

PR #507, opened 2024-01-27
docs: add URLs for most repositories in the docs

link2xt:
This makes git repository overview more usable
as it is possible to follow the link directly
instead of searching for the repo on GitHub.


david415:
hello @link2xt thanks for caring about katzenpost but your pull request is deeply flawed. we are currently using the monorepo... this same repo and not all the git repos listed in your PR... which were marked as deprecated.

PR #512, opened 2024-02-04
Add plugin Parameters to CBORPlugin using tagged types

### LINKED ISSUE #481: cborplugin system no longer supports plugin Parameters
previously, the mix server plugin system provided for cborplugin.ServerPlugin to expose parameters via HTTP requests to URL path /parameters.

This functionality is no longer present as the cborplugin system speaks CBOR over a unix socket, but does not provide a way to request parameters to publish in the PKI.
In order to provide this functionality, the plugin should be able to receive a request for parameters and provide a response.

For example, https://github.com/katzenpost/katzenpost/tree/add_cborplugin_parameters provides an example of using CBOR tags so that the plugin may accept and return serialised CBOR corresponding to different types of reqeusts.

https://github.com/katzenpost/katzenpost/blob/f787de19894f90db1648075a103dc8eade144a47/server/cborplugin/client.go#L41 (TagSet is defined)
https://github.com/katzenpost/katzenpost/blob/f787de19894f90db1648075a103dc8eade144a47/server/cborplugin/socket.go#L114 (socket reader uses TagSet)
https://github.com/katzenpost/katzenpost/blob/f787de19894f90db1648075a103dc8eade144a47/server/cborplugin/socket.go#L137 (socket writer uses TagSet)

mixmasala:
This re-adds support for a plugin to expose Parameters in the PKI by way of tagged types.

Previously, server plugins spoke HTTP and supported requests to /parameters endpoint, which enabled a plugin to set custom values in the mix server's MixDescriptor.
This was broken when the SocketIO CBOR-over-unix-socket replaced the CBOR-over-HTTP interface, which caused the reunion server plugin to no longer work.

This PR adds new type ParametersRequest and updates existing plugins. Also adds unit test for GetParameters. Fixes #481, progress on #482.


david415:
Hey I brought this topic up several times over the years: tags are harder to use than embedded structs;
Remember when I posted this question to the go-nuts google groups?

https://groups.google.com/g/golang-nuts/c/TXGRL2nmDxk

I think we should instead, switch to using embedded structs like we currently do, everywhere.

david415 (changes requested):
remove use of CBOR tags, replace with nested structs

ioerror:
@mixmasala Can you resolve the conflicts and then re-request a review?

PR #520, opened 2024-02-22
Gateway/Service nodes instead of Provider nodes

### LINKED ISSUE #545: Gateway originated decoy
After we solve for https://github.com/katzenpost/katzenpost/issues/544 then we must make Gateways send decoy loops; and it must be tuned properly as per @evainfeld's technical report on tuning.

david415:
We wish to do away with the old notion of Providers being a specialized type of mix node and instead create two roles that are specialized mix nodes:

Gateway nodes' only task is to allow mixnet clients to connect and interact with the mixnet.
Service nodes' only task is to allow service plugins to run so that the mixnet can interact with them.

This changeset also

* removes the "thwack" management socket from the mix server because it is obsolete code that is totally unused with our current use cases for Katzenpost.
* ALL mix nodes including gateway AND service nodes now send heartbeat traffic HOWEVER the gateway nodes do so at the rate expressed by LambdaG in the PKI doc. Whereas service nodes and mix nodes use LambdaM.
* LambdaG comes from Dr. Eva Infeld's Katzenpost mixnet tuning document and we calculate it via n*ln(n) where n is the number of mix nodes in the first topology layer. We're doing this calculation in the dirauth and then it's value is propagated to everyone else via the LambdaG field in the PKI document. It's only acted upon by the Gateway nodes, hence the "G" in LambdaG.
* We add `IsGatewayNode` and `IsServiceNode` fields to the `MixDescriptor` type
* The directory authority enforces whether or not a node is able to register itself as a gateway or service node or mix node.
* Our PQ Noise based wire protocol still enforces which layer of nodes can connect to which layer. The last layer of mix nodes is able to connect to the gateway nodes AND the service nodes. Likewise the Service nodes layer AND the Gateway nodes layer allow connections from the last layer of mixes. Therefore some next-layer or previous-layer may in fact return two layers with our new topology changes that are accurately reflected in this changeset.

Fixes #545 #544


ioerror:
Please resolve the outstanding conflicts to ensure a timely review process.

david415:
Currently, the docker mixnet does not ping, I've documented the gateway logs and ping logs in this ticket:
https://github.com/katzenpost/katzenpost/issues/550

amitbasuri (approved):
LGTM

PR #523, opened 2024-03-03
Fixup kem usage

david415:
Configure the wire KEM scheme via TOML configuration parameter for client, mix node and dirauth node!


david415:
this doesn't quite work yet. needs a little bit more work.

ioerror (approved):
LGTM

PR #526, opened 2024-03-19
Propagate decoy stats to dirauths

david415:


evainfeld (approved):
Looks good. I would like to iterate on this. For example, are we in agreement that we should be indexing by connections?

ioerror (approved):
LGTM

Issue #527, opened 2024-03-21
CGO error when consuming Katzenpost from 3rd party repo/code


david415:
Currently main branch of Katzenpost is broken even though all the tests pass because 3rd party consumers of Katzenpost cannot use Katzenpost; broken broken broken.

You can easily demonstrate this error for yourself:
```bash
git clone https://github.com/katzenpost/3rd_party_clients.git
cd 3rd_party_clients/ping
go build
```

The above results in this error message:

```
github.com/katzenpost/katzenpost/sphincsplus/ref: invalid flag in #cgo CFLAGS: -DPARAMS=sphincs-shake-256f
```

david415:
fixed

david415:
THIS BUG IS FIXED BY THIS PR https://github.com/katzenpost/katzenpost/pull/528

PR #528, opened 2024-03-21
use latest hpqc

david415:


david415:
THIS PR FIXED https://github.com/katzenpost/katzenpost/issues/527

PR #537, opened 2024-04-13
Fix docker makefile

leif:


mixmasala (commented):
thank you for fixing this

mixmasala (approved):

Issue #371, opened 2023-08-08
Noise message padding for all mixnet commands


david415:
We'd like to pad all Noise protocol messages that the mixnet uses such that they are all the same size as the Message command containing Sphinx packets. The outlier of course is the Consensus command which contains very large PKI documents of around ~200k for the docker mixnet.

These are the wire protocol commands:
https://github.com/katzenpost/katzenpost/tree/main/core/wire/commands

The basic idea is that we want each of these wire commands to serialize to the same size so that a passive network observer cannot determine which command is being sent or received. This task should be done along with unit tests that show that all the wire protocol commands are all the same size. The tests should serialize each command and then check that they are all the same size.

amitbasuri:
Hi @david415 I've reviewed the implementation of command serialization and noticed that we currently handle different command lengths specifically tailored to each command's requirements. 

Now to prevent traffic analysis by standardizing command lengths, I propose two approaches, considering the unique size of the Consensus command:

#### Proposal 1: Uniform Command Size
- **Uniform Size**: Adopt a `maxCommandSize` across all commands, set at the size needed for the largest command (potentially 210 KB).
- **Zero Padding**: Implement zero-padding for all commands up to the `maxCommandSize`.

#### Proposal 2: Special Handling for Consensus
- **Reduced Size for Common Commands**: Set a smaller `maxCommandSize` for regular commands (e.g., 100 KB).
- **Separate Treatment for Consensus**: Allow the Consensus command to exceed this size, using segmentation or a higher limit.

**Discussion Points:**
- Which approach better balances security with operational efficiency?

Looking forward to your thoughts on which strategy might be more effective for our needs.

david415:
@amitbasuri Good summary. I'd like us to implement Proposal 2.

david415:
fixed forever!

PR #548, opened 2024-04-26
Implement Message Padding for Mixnet Commands

amitbasuri:
Issue: https://github.com/katzenpost/katzenpost/issues/371

This PR introduces uniform padding for all Noise protocol messages used in the mixnet, aligning them to the maximum size of the largest command, the Consensus command. The padding ensures that all messages serialize to the same size, preventing traffic analysis by obscuring the command types from passive network observers. The changes include modifications to the serialization functions for each command to apply zero-padding up to the specified maximum size, and unit tests have been added to verify that all commands meet the new size requirement.


david415:
Very good start. I'm sorry I didn't explain all the details in the ticket. Here I will share more details.

The following commands are not sent over the mixnet, instead they are sent by mix nodes to the dirauths directly or they are sent among the dirauths:

GetVote
PostDescriptor
PostDescriptorStatus
Reveal
RevealStatus
Vote
VoteStatus
Cert
CertStatus
Sig
SigStatus

We do not need to pad the above set of commands.
We want to pad these commands:

GetConsensus
NoOp
Disconnect
SendPacket
RetrieveMessage
MessageACK
Message
MessageEmpty

The largest command among the above will be the Message command because it contains a Sphinx packet. The size of the sphinx packet depends on the configured Sphinx Geometry for that mixnet.

The unit test must check that the commands are serialized to the same size. Currently it's only checking that the padding are all zero bytes. It should check that the padding bytes are zero but you should use this helper function for that:

https://github.com/katzenpost/hpqc/blob/main/util/ctIsZero.go

david415:
```
go test -v -run TestCommandVectors
=== RUN   TestCommandVectors
--- PASS: TestCommandVectors (0.00s)
PASS
ok  	github.com/katzenpost/katzenpost/core/wire/commands	0.002s
```

These tests should fail given that the test vectors were not updated.

amitbasuri:
@david415 

1) For TestCommandVectors I am only asserting without the padding bytes `sendPacketBytes[:len(sendPacketCommand)]`
```go
	sendPacketBytes := sendPacket.ToBytes()
	assert.Equal(sendPacketBytes[:len(sendPacketCommand)], sendPacketCommand)
	assert.True(util.CtIsZero(sendPacketBytes[len(sendPacketCommand):]), "SendPacket: ToBytes() padding must be zero")
```

2) I passed `*geo.Geometry` to client to resolve `TestClient`

david415:
> @david415
> 
>     1. For TestCommandVectors I am only asserting without the padding bytes `sendPacketBytes[:len(sendPacketCommand)]`
> 
> 
> ```go
> 	sendPacketBytes := sendPacket.ToBytes()
> 	assert.Equal(sendPacketBytes[:len(sendPacketCommand)], sendPacketCommand)
> 	assert.True(util.CtIsZero(sendPacketBytes[len(sendPacketCommand):]), "SendPacket: ToBytes() padding must be zero")
> ```
> 
>     2. I passed `*geo.Geometry` to client to resolve `TestClient`

All the integration tests are failing.

david415:
Among the set of commands we want to pad:

GetConsensus
NoOp
Disconnect
SendPacket
RetrieveMessage
MessageACK
Message
MessageEmpty

I just now remembered that we have two directions. From client to server the `SendPacket` command is the biggest because it contains an entire Sphinx packet. What size is that? geo.PacketLength https://github.com/katzenpost/katzenpost/blob/main/core/sphinx/geo/geo.go#L44

But from server to client the biggest (discounting Consensus of course) is `Message` or `MessageAck`. However these are both smaller than `SendPacket` because they only contain part of the Sphinx payload (without the added SURB).

Therefore, we need two padding sizes... and two established lists of commands, egress vs ingress commands. The idea is that all the commands flowing from client to server will be padded to the same size, size A. And the commands flowing from server to client will be padded to size B.

david415:
The `GetConsensus` command is a special case because we only want to pad it when it's sent over the mixnet. We don't want to pad it when sending to the dirauths.

david415 (approved):

PR #558, opened 2024-05-10
eradicate eapache channels library, it's deprecated

david415:
Remove a deprecated library called the eapache channel library.
Previously we were using it's InfiniteChannel... and in this PR i'm replacing it with a buffered channel. buffered to length 1000 is more than enough to handle bursty traffic loads I'm guessing. Likewise 1000 wouldn't take that much system memory.


threebithacker (approved):
Looks good to me

PR #561, opened 2024-05-11
add work-in-progress SEDA pipeline selftest

### LINKED ISSUE #559: mix server: SEDA pipeline self-test
Introduction
===========

Staged Event Driven Architecture, SEDA has influenced the design of our mix server. Internally the mix server is compose of 4 thread pools connected together using queues and AQMs, Active Queue Management algorithms.

1. Input packet worker thread pool
2. crypto worker thread pool
3. mixing scheduler thread
4. output packet worker thread pool

The input packet worker thread is responsible to handling our wire protocol and receiving packets and sending
them off to the next element in the pipeline, in this case the cryptoworker thread pool. The cryptoworker thread pool is responsible for decrypting the Sphinx packets. The mixing scheduler delays packets as per their specified Sphinx delay routing command. The output packet worker thread pool simply sends each Sphinx packet to the next hop, the next mix node in it's route through the mix network.

Read about SEDA: here https://www.sosp.org/2001/papers/welsh.pdf
And here's a brief wikipedia article on AQMs: https://en.wikipedia.org/wiki/Active_queue_management

The advantage that SEDA gives us over a simpler design that does all the packet process

The Problem Statement
====================

Since we have a pipeline that uses a SEDA design, it therefore necessarily requires tuning. Each of the AQMs in the pipeline has a threshold delay which if exceeded causes it to drop packets; in other words, if there is not enough computational resources, the thread pool won't react in time and the AQM that is attached may decide to drop packets.
We need to be able to set each of the threshold values for each of the AQMs.

The Solution
==========

I suspect a "self-test mode" for the mix server would be useful; it would perform a self-test operation in order to discover the threshold values for each of the AQMs. Now, of course the threshold values are specific to the hardware being used. But another observation is that they are also specific to the PQ Noise AND Sphinx cryptographic primitives being used as well since those are what slows it down.

A command line flag could cause the mix server to run in self-test mode. Essentially what we want to do is compose a bunch of one-hop Sphinx packets and queue them up. Then we can perform tests by sending them into the mix server's input packet worker... and having them flow throughout the pipeline until they are emitted in the output packet worker.
I would initially start off with all of the thresholds set very high so that they don't cause any packet loss. Then you could try tweaking one threshold at a time, gradually lowering it's timeout value until it causes packet loss.

The standard golang net.Conn interface is useful here because we can implement any sort of connection and put our wire protocol on top of it. Therefore... for testing, we don't want to use a TCP connection. Instead use a net.Pipe! Or perhaps something else if you prefer. We can certainly make our own interfaces if it makes things easier.

Are there a couple problems with this plan? Firstly, the mix server does not yet have any central events channel for reporting packet loss. We should create that in order to detect packet loss from any of the AQMs in the pipeline. Secondly, the AQMs aren't designed to handle a dynamically changing threshold value; they are all static values that come from the TOML configuration file. So we maybe need to add some code to each AQM that lets it reset it's timer to the new deadline if it ever changes.

For this ticket we are mostly concerned with deriving optimal settings for these mix server configuration fields:

https://github.com/katzenpost/katzenpost/blob/main/server/config/config.go#L203-L235

however we might want to do some testing to figure out the best values for these as well:

https://github.com/katzenpost/katzenpost/blob/main/server/config/config.go#L237-L247

david415:
fixes #559


ioerror:
Per discussion with @david415, this PR will be closed.

Issue #562, opened 2024-05-11
WarpedEpoch is probably wrong


david415:
Here we see WarpedEpoch is set incorrectly here:
https://github.com/katzenpost/katzenpost/blob/main/docker/Makefile#L60

```
ldflags="-buildid= -X github.com/katzenpost/katzenpost/core/epochtime.WarpedEpoch=${warped} -X github.com/katzenpost/katzenpost/server/internal/pki.WarpedEpoch=${warped} -X github.com/katzenpost/katzenpost/minclient/pki.WarpedEpoch=${warped}"
```

This directory exists:

server/internal/pki

this directory does not exist:

minclient/pki

whereas:

minclient/pki.go exists

and:

server/internal/pki/pki.go exists

Therefore the above 3 different paths CANNOT ALL BE CORRECT.

PR #566, opened 2024-05-15
remove some WarpedEpoch-related cruft

leif:
we used to need to set and check WarpedEpoch in more places but now it is only used in core/epochtime/time.go so we only need to set it there.

fixes https://github.com/katzenpost/katzenpost/issues/562


mixmasala (approved):
thanks for cleaning this up

PR #572, opened 2024-05-16
560 Refactor Signature scheme

amitbasuri:


david415:
@amitbasuri be sure to merge main into your branch so that it's up to date ;-)

david415:
@amitbasuri Looks good so far. The signature KEM scheme needs to be a field in the authority client config -> https://github.com/katzenpost/katzenpost/blob/main/authority/voting/client/client.go#L71-L90

authority/voting/client/client.go
```diff
@@ -162,17 +163,22 @@ func (p *connector) initSession(ctx context.Context, doneCh <-chan interface{},
 
 	// Initialize the wire protocol session.
 	var ad []byte
+	var pkiSignatureScheme sign.Scheme
 	if signingKey != nil {
 		keyHash := hash.Sum256From(signingKey)
 		ad = keyHash[:]
+		pkiSignatureScheme = signSchemes.ByName(signingKey.Scheme().Name())
+	} else {
+		pkiSignatureScheme = signSchemes.ByName("Ed25519 Sphincs+") // todo fix me
```
david415, line 172: it should obtain the signature scheme from the config file

david415: if signingKey is allowed to be nil then you should not be deriving the scheme from it. instead get the signing scheme from the config.

david415 (changes requested):
I think the only thing missing is to add the PKISignatureScheme field to the voting authority client config. That we must do and then I want to merge this as soon as possible.

PR #580, opened 2024-05-17
add goreleaser

david415:


ioerror (approved):
The windows build here will probably need additional work and similarly the MacOS build too. It is a fine start.

ioerror (changes requested):
It looks like the release.yml has a syntax error and needs to be fixed: https://github.com/katzenpost/katzenpost/actions/runs/9262989116/workflow

(run `yamllint` as a pre-commit hook or manually)

amitbasuri (approved):

ioerror:
The overall goreleaser work is a fine start but the actual CI configuration yaml file has a syntax error on line 13 of `.github/workflows/release.yml`. That syntax error needs to be fixed for this this to use the goreleaser configuration.

Issue #581, opened 2024-05-18
Add CI testing matrix with various os/arch permutations


david415:
We want to run the unit tests (NOT the docker tests) via the github CI testing matrix that specifies how to run the unit tests on each arch/os permutation, there are a great many.

There also might be some useful testing tricks we can learn from the highctidh project:
https://codeberg.org/vula/highctidh

PR #583, opened 2024-05-20
add multiple os/arch for github ci test

amitbasuri:
Resolves https://github.com/katzenpost/katzenpost/issues/581


david415:
Lets remove the centos. Also let's not test with ubuntu and debian... just one of them.
It would be cool if we could have darwin and windows working here.
We need the tests to all pass.

amitbasuri:
@david415  The `hpqc` repo fails CI test for macos/windows  ref https://github.com/katzenpost/hpqc/pull/17
Which is causing `katzenpost` test to fail.

david415:
> @david415 The `hpqc` repo fails CI test for macos/windows ref [katzenpost/hpqc#17](https://github.com/katzenpost/hpqc/pull/17) Which is causing `katzenpost` test to fail.

Thank you @amitbasuri this is as far as we can get with this ticket now. The rest will have to be dealth with inside hpqc.

Issue #574, opened 2024-05-17
Add mix server prometheus metrics


david415:
We recently merged this PR https://github.com/katzenpost/katzenpost/pull/558 which got rid of the eapache channels library however we replaced the InfiniteChannel with a buffered channel of size 1000. If we add some more Prometheus metrics to the Katzenpost mix server then we will be able to determine how much of these buffered channels are used during the most busty traffic changes. This will help us set the size of these channels to be much smaller and could result in a memory saving of a few megabytes, possibly dozens of megabytes when we add up all the goroutines using these buffered channels.

The prometheus metrics that we want to emit are a gauge or histogram... not a counter because the number of items in a buffered channel can go up or down at any time. len(ch) to get the number of elements:

https://stackoverflow.com/questions/13003749/number-of-elements-in-a-channel

PR #584, opened 2024-05-20
add prometheus Gauge Channel Length

amitbasuri:
Resolves https://github.com/katzenpost/katzenpost/issues/574

### DIFF
--- a/server/internal/gateway/gateway.go
+++ b/server/internal/gateway/gateway.go
@@ -311,6 +311,18 @@ func New(glue glue.Glue) (glue.Gateway, error) {
 		p.Go(p.worker)
 	}
 
+	// monitor channel length
+	go p.monitorChannelLen()
+
 	isOk = true
 	return p, nil
+}
+
+func (p *gateway) monitorChannelLen() {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	for range ticker.C {
+		instrument.GaugeChannelLength("server.gateway.ch", len(p.ch))
+	}
 }
--- a/server/internal/gateway/gateway.go
+++ b/server/internal/gateway/gateway.go
@@ -312,7 +312,7 @@ func New(glue glue.Glue) (glue.Gateway, error) {
 	}
 
 	// monitor channel length
-	instrument.MonitorChannelLen("server.gateway.ch", p.ch)
+	instrument.MonitorChannelLen("server.gateway.ch", p.HaltCh(), p.ch)
 
 	isOk = true
 	return p, nil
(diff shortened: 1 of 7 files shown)


server/internal/gateway/gateway.go
```diff
@@ -311,6 +311,18 @@ func New(glue glue.Glue) (glue.Gateway, error) {
...
+
 	isOk = true
 	return p, nil
 }
+
+func (p *gateway) monitorChannelLen() {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	for range ticker.C {
```
david415, line 325: We've got a worker type and this `gateway` type is composite with the `Worker` type:
https://github.com/katzenpost/katzenpost/blob/main/server/internal/gateway/gateway.go#L38
https://github.com/katzenpost/katzenpost/blob/main/core/worker/worker.go#L23

And so... a very common pattern we use inside the mix server is to maintain a server shutdown code path:

```golang
select {
        case <-p.HaltCh():
                return
        case <-timer.C:
                // blah blah
}
```

I'm not saying we need to do this here necessarily unless the mix server no longer shutdown down cleanly.

server/internal/instrument/prometheus.go
```diff
@@ -167,6 +169,31 @@ var (
 	)
 )
 
+var monitoredChannels = struct {
+	sync.Mutex
+	channels map[string]chan interface{}
```
david415, line 174: there's no need to create a struct type and store channels in a map.

server/internal/instrument/prometheus.go
```diff
@@ -167,6 +169,31 @@ var (
...
+}{
+	channels: make(map[string]chan interface{}),
+}
+
+func startChannelLenMonitor() {
```
david415, line 179: this isn't ideal. we do not need nested for loops or mutex contention or a map storing things.

david415: i fixed this by pushing another commit on this branch

david415:
I pushed a commit to this branch that simplifies the prometheus monitoring of buffered channels:

```
func MonitorChannelLen(name string, haltCh <-chan interface{}, ch chan interface{}) {
	go doMonitorChannelLen(name, haltCh, ch)
}

func doMonitorChannelLen(name string, haltCh <-chan interface{}, ch chan interface{}) {
	ticker := time.NewTicker(time.Second)
	defer ticker.Stop()
	for {
		select {
		case <-haltCh:
			return
		case <-ticker.C:
			channelUsage.With(prometheus.Labels{"channel_name": name}).Set(float64(len(ch)))
		}
	}
}
```

amitbasuri:
@david415 yes now it looks better

david415 (approved):

PR #591, opened 2024-05-24
Fix packet dropping bug

### LINKED ISSUE #568: high load mix server drops Sphinx packets with bad mac err
during high load for some reason in our testnet the mix server drops lots of Sphinx packets with bad mac err even though the mac being bad is seemingly not possible. the goal of this ticket is to carefully audit the mix server code to find and fix the bug where upon high load it manifests by dropping packets.

errs like this:

```
01:59:26.390 DEBU crypto:1: Dropping packet: 1978 (sphinx: invalid packet, MAC mismatch)
01:59:26.493 DEBU crypto:0: Dropping packet: 1980 (sphinx: invalid packet, MAC mismatch)
```

david415:
fixes #568 a bug where the mix node drops packets at every epoch boundary, finally fixed after all these years.


server/internal/cryptoworker/crypto_worker.go
```diff
@@ -103,10 +103,10 @@ func (w *Worker) doUnwrap(pkt *packet.Packet) error {
 
 		w.log.Debugf("Packet: %v (Unwrap took: %v)", pkt.ID, unwrapAt.Sub(startAt))
 
+		lastErr = err
```
amitbasuri, line 106: How about if we just return error for the below conditions 

```go
		if err = pkt.Set(payload, cmds); err != nil {
			return fmt.Errorf("crypto: Failed to set payload: %w", err)
		}

		// Check for replayed packets.
		if k.IsReplay(tag) {
			// The packet decrypted successfully, the MAC was valid, and the
			// tag was seen before, therefore drop the packet as a replay.
			instrument.PacketsReplayed()
			return errors.New("crypto: Packet is a replay")
		}

david415: that wouldn't do the same thing because we only want to return the error after the for loop is done.

ioerror:
I have merged main into this branch and if the tests pass, I am satisfied that this is ready to merge.

ioerror:
There is a bug in the completely unrelated derive secret hpqc that is causing test_e2e_catshadow to fail. This has been broken on main for a while - so i think that this should not block merging the fix for this very long standing issue. The issue causing an intermittent failure in the `test_e2e_catshadow` test is in #615.

ioerror (approved):

PR #594, opened 2024-05-25
server: MUCH better error message!!!

david415:
the debug log error message is confusing: "Out of epoch" .
here we fix this by changing it to an error message which explains the problem.


ioerror (approved):
LGTM

ioerror:
We really need to address the hanging test that never runs. This is a CI configuration issue that is unrelated to this change.

PR #599, opened 2024-05-28
docker: make compatible with docker v2

leif:
on systems with a new enough version of the Python requests module it is no longer possible to use the (deprecated) standalone docker-compose v1.

with this commit we will now use docker compose v2 (the "compose" subcommand to docker) if it is installed. we are calling its executable directly, which can drive either podman or docker.

this commit also makes the shell target create the network directory if it does not exist. (note that we still run mkdir -p in the docker_compose_yml target separately rather than depending on the target that creates it because depending on that causes the config to be regenerated every time the directory is modified.)


mixmasala (approved):
this works for me in my development environment

PR #600, opened 2024-05-28
doubleratchet: make upgrading from pre-nike doubleratchet work

mixmasala:
fixes #598

PR #608, opened 2024-05-29
Fix ratchet

david415:


ioerror:
What's up with #600? Do you think you guys could work together to fix things rather than competing requests with cherry-picking?

david415:
> What's up with #600? Do you think you guys could work together to fix things rather than competing requests with cherry-picking?

in #600 they didn't bother to fix hpqc and thus selected a ratchet NIKE which doesn't work on all platforms?

ioerror:
Please use highctidh v1.0.2024052900 - it was tagged just to solve your problem.

david415:
> Please use highctidh v1.0.2024052900 - it was tagged just to solve your problem.

thanks! should be using that tag now.

ioerror:
Please update this to use tag  v1.0.2024053000.

ioerror:
To fix the CI, I think you'll want to close and reopen this pull request.

ioerror:
In discussion with @david415; this will be closed (and a new PR opened) to try to shake the CI bug out.

PR #609, opened 2024-05-29
BUG FIX: Remove prometheus goroutines for monitoring buf chans

david415:
The extra goroutines used for the Prometheus gauges of buffered channel usage are causing high mix server load.


ioerror (approved):
These changes resolve a bug which impacts all nodes with higher CPU load. This changes the ability to use Prometheus to graph data. There is no test that catches this bug.

PR #613, opened 2024-05-30
Fix copyright notice

david415:


ioerror (approved):
This is good; please ensure all files retain their copyright/copyleft notices. Probably it should be a policy that every file has such a header and additionally an authors file could collect them all together.

Issue #615, opened 2024-05-31
csidh.DeriveSecret failed


ioerror:
It seems that the latest hpqc has a problem using nobs but sometimes it does work. Usually it fails:
```
10:39:33.041 DEBU minclient/conn:gateway1: Received MessageEmpty: 30
panic: csidh.DeriveSecret failed!
```

david415:
this bug has been fixed

PR #617, opened 2024-05-31
genconfig: default ratchet to CTIDH512-X25519

david415:


ioerror:
This is related to #615

ioerror:
If possible, it would be nice to have this test in parallel with the nobs csidh test - so we can see if one fails, if the other fails in the parallel run. I suspect the answer is that the nobs interface is somehow not as expected.

ioerror (changes requested):
Could you please add a second test that tests the nobs usage in parallel to show that hpqc enables a functional catshadow with either parameter?

It seems that the hpqc nobs interface is not returning the same or similarly acceptable values as it used to return when the test previously passed.

david415:
> Could you please add a second test that tests the nobs usage in parallel to show that hpqc enables a functional catshadow with either parameter?
> 
> It seems that the hpqc nobs interface is not returning the same or similarly acceptable values as it used to return when the test previously passed.

no. adding the second test is actually too much docker cpu time and it will take too long. we already run each test twice because of the push vs pull in the ci matrix. we can test the separately... like i'm doing now with the latest change to use the nobs csidh hybrid nike.

ioerror (approved):
Good enough for this iteration, thanks.

PR #624, opened 2024-06-02
testnet: cache build artifacts

leif:
With this PR, the /go and /root/.cache directories are mounted from outside the containers (in ./docker/cache) which makes building *much* faster. This also means we don't need to maintain a separate go_mod docker/podman image anymore; we just use the base image with distro packages now.

The need for reducing the time spent building became especially noticeable since we added highctidh, which prior to this PR is making each of our many binaries which link against it take minutes to build. With this build cache, all binaries after the first one build in mere seconds again.

The pipe syntax in the Makefile designates "order-only-prerequisites", so that targets which depend on $(cache_dir) won't be re-run just because the mtime of the cache dir is newer than they are. See
https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html for details.

This PR also:
* removes the katzencat make target (as katzencat hasn't been merged yet)
* makes the docker-mixnet ci workflow use the "wait" target (which uses the "fetch" command masala wrote last year) to find out when the mixnet is ready, instead of sleeping four minutes

### DIFF
--- a/.github/workflows/darwin.yml
+++ b/.github/workflows/darwin.yml
@@ -1,5 +1,5 @@
 name: darwin unit tests
-on: [push, pull_request]
+on: [pull_request]
 jobs:
   test:
     strategy:
@@ -7,7 +7,7 @@ jobs:
       matrix:
         os: [macos-latest]
         arch: [amd64, arm64]
-        go-version: [ "1.21.x", "1.22.x"]
+        go-version: ["1.22.x"]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Install Go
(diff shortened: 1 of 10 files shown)


ioerror:
I kicked off the failed test to see if it will indeed pass.

ioerror (approved):
Assuming that the CI passes completely, the changes seem a strict improvement - especially the removal of the awful sleep 240 which was terrorizing everyone with a delay that was often far far too long.

ioerror:
The `test_e2e_catshadow` test is unreasonably slow. 42 minutes and counting....

david415 (approved):

ioerror:
I am confused as to why the ` test_e2e_catshadow` test that keeps failing is failing every time. This implies that this PR someone makes things worse. Does the current test not test the same code paths or is the previously passing test is a red herring? Unrelated, I see that someone git tip only has two tests and one of them fails due to cgo, so that's another concern...

PR #637, opened 2024-06-12
add crypto_worker_test

amitbasuri:


david415:
i think we want to close this since all the tests fail and currently crypto worker is well tested.

PR #638, opened 2024-06-12
server: rearrange code to make routing logic testable

david415:


amitbasuri:
Hi @david415 I recommend using a library like [Counterfeiter](https://github.com/maxbrunsfeld/counterfeiter) for generating mocks to streamline testing. It automates mock generation, ensures consistency, and allows us to focus more on the test logic rather than manual mock management, improving our development workflow and reducing code duplication.

david415:
> Hi @david415 I recommend using a library like [Counterfeiter](https://github.com/maxbrunsfeld/counterfeiter) for generating mocks to streamline testing. It automates mock generation, ensures consistency, and allows us to focus more on the test logic rather than manual mock management, improving our development workflow and reducing code duplication.

Here my manual mocked objects are keeping track of counters which I use to determine the routing outcome for each test case.

PR #651, opened 2024-06-15
make it possible to upgrade to the nike doubleratchet

leif:
A refactor of doubleratchet recently landed which makes it use a generic NIKE from hpqc instead of combining x25519 and CSIDH itself as it was doing before.

The [expectation](https://github.com/katzenpost/katzenpost/pull/250#issuecomment-1455304465) was that it would be backwards compatible when the correct hybrid NIKE was selected, but it was not tested and manual testing later revealed that it did not work with the CSIDH nike at all. After that was fixed in #656, we discovered that it was still not backwards compatible.

This merge:
* adds automated upgrade testing using state files from an old version to avoid this happening in the future
* modifies doubleratchet to use the same header layout as it did in the previous version, which makes these tests pass. the actual fix is here: https://github.com/katzenpost/katzenpost/pull/651/commits/6aae522ea2ca4d5d15f533b13f5a691280565708
* includes git history from a katzenpost-v0.0.30-based branch where these tests were backported to, in order to create new client state files with the old version for testing upgrades
* improves logging in various places
* makes memspool call doFlush after every mutation. this is not very efficient but without it i would need to wait 30 seconds after each write when creating these automated memspool tests. memspool can't be gone soon enough; hopefully with the changes on this branch we can provide users of it a working migration path when one is ready.
* adds docker Makefile targets for starting, stopping, and restarting memspool using its management socket, and makes other improvements to the docker Makefile
* removes the Makefile behavior of stopping affected containers automatically when rebuilding binaries

At https://github.com/katzenpost/katzenpost/pull/649/files is the v0.0.30-backported version of the tests in this PR where the v30 state files for testing were created.

### DIFF
--- a/memspool/server/cmd/memspool/main.go
+++ b/memspool/server/cmd/memspool/main.go
@@ -59,7 +59,7 @@ func main() {
 	}
 
 	// Log to a file.
-	logFile := path.Join(logDir, fmt.Sprintf("memspool.log"))
+	logFile := path.Join(logDir, fmt.Sprintf("memspool.%d.log", os.Getpid()))
 	logBackend, err := log.New(logFile, logLevel, false)
 	if err != nil {
 		panic(err)


david415:
your old state files are corrupt:
```
 --- FAIL: TestUpgradeResume_1 (0.08s)
panic: RecvRatchetPublic is all zeros [recovered]
	panic: RecvRatchetPublic is all zeros
```

also using an integration test is the wrong approach.

mixmasala (changes requested):
That's a tremendous amount of work! Well done spotting the key serialization bug :-).
And now we have a framework for testing interopability with previous versions of our software too.

A few small items I noticed:

* do not log spool ID's on ERROR loglevel messages
* keep the PID logfile naming, in case of multiple workers
* is there another way to get memspool to sync cleanly? are you sure it is losing messages?

.github/workflows/docker-mixnet.yml
```diff
@@ -74,6 +74,9 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Install socat, for talking to management socket
+        run: sudo apt update && sudo apt install -y socat
```
david415, line 78: ah okay so this installs socat so that you can shell into the docker image and use it from there? or can you socat from the host?

catshadow/operations.go
```diff
@@ -25,6 +25,9 @@ import (
 	"github.com/katzenpost/katzenpost/memspool/client"
 )
 
+type opNop struct {
```
david415, line 28: what's the Nop used for?

david415 (approved):
Looks good to me. I left a couple of questions but I didn't see anything here that would cause me to request changes.

mixmasala (approved):

PR #654, opened 2024-06-15
Fix packet dropping bug with counterfieter lib

amitbasuri:
Hi @david415 I wanted see how the mockgen lib will workout for `Glue interface` , so I tried it for `TestRoutePacket`


mixmasala (approved):

david415:
obviously we can't merge this because all the tests fail. closing pr

PR #656, opened 2024-06-16
ratchet_test.go: use NOBS_CSIDH512X25519

leif:
This changes the doubleratchet test's nike to NOBS_CSIDH512X25519, and upgrades to the latest hpqc with https://github.com/katzenpost/hpqc/pull/30 which fixes the bug which made doubleratchet unable to use it.


mixmasala (changes requested):
should hpqc be tagged?

mixmasala (approved):

PR #661, opened 2024-06-17
add quic listeners

mixmasala:
Adds QUIC transport to authority and server


server/internal/pki/pki.go
```diff
@@ -696,26 +695,16 @@ func New(glue glue.Glue) (glue.PKI, error) {
 	}
 
 	var err error
-	if glue.Config().Server.OnlyAdvertiseAltAddresses {
+	if len(glue.Config().Server.OnlyAdvertiseAddresses) > 0 {
 		p.descAddrMap = make(map[string][]string)
+		// XXX: parse each address and update descAddrMap
+		panic("NotImplemented")
```
david415, line 701: what isn't being implemented here?

david415:
does quic-go have any advantages over using the quic in the golang standard library?
I like the new url format for the addresses.
shall we merge this now?

david415 (approved):

mixmasala:
I took a quic look at golang.org/x/net/quic and it looked like the interfaces are very similar to those of quic-go.
There is a good chance that changing the import path might just work. I'm not sure how much code these implementations share.

david415 (approved):

PR #670, opened 2024-07-01
docker instructions for review

jdormansteele:
This is the first draft of the docker test mixnet procedures, submitted for team review and response.

There are two documents:

- docker.pdf, which is a reader-friendly rendering of the content. Comments from me containing questions or requests for assistance are displayed as footnotes. This document cannot be edited.
- docker.md, which you can edit. The comments from me, the same ones as in the PDF, are embedded here as HTML comments. Please review the content for accuracy, and make corrections, additions, or other changes as needed. You can also respond to me verbosely in the document using HTML comments.

In this draft, please focus on the procedures, which are basically complete, but also note that there is a topology diagram and a table of mixnet components requiring your review and corrections/suggestions.

When the team is done reviewing this draft, I will incorporate the changes into my master document and generate new drafts for you, repeating this cycle as many times as needed to have a document ready for publication.


david415:
Looks good to me. How can we convert between markdown and the pdf?

jdormansteele:
Production of the PDF is a one-way process based on ANT transform of the original XML. I included that just a clarification in case the markdown is confusing.

david415:
Table 1.2 needs to not say anything about databases. No Spool DB and no User DB. Those are internal details that do not make sense to discuss in the docker document. We can maybe discuss them in the mix server documentation if/when you get around to writing that.

Table 1.2 needs to show a Gateway node. Every Katzenpost mixnet has to have at least one Gateway node. The Gateway node is where the Spool DB and User DB reside... although we do not want to discuss those internal details in this document ;-) they certainly do not reside on the Service node.

Table 1.2 needs to elaborate on the list of services offered by the Service Node.
Our current docker mixnet is running several services. List them all, please.

david415 (changes requested):
Table 1.2 needs fixing

david415:
the node types are:

mix node
service node
gateway node


we no longer have "providers" and also we never had "service providers".

jdormansteele:
> Looks good to me. How can we convert between markdown and the pdf?

In short, please submit revisions *in* the markdown.

jdormansteele:
> Table 1.2 needs to not say anything about databases. No Spool DB and no User DB. Those are internal details that do not make sense to discuss in the docker document. We can maybe discuss them in the mix server documentation if/when you get around to writing that.
> 
> Table 1.2 needs to show a Gateway node. Every Katzenpost mixnet has to have at least one Gateway node. The Gateway node is where the Spool DB and User DB reside... although we do not want to discuss those internal details in this document ;-) they certainly do not reside on the Service node.
> 
> [DWR] For the Gateway node, what are the values for Identifier, IP, and Port?
> 
> Table 1.2 needs to elaborate on the list of services offered by the Service Node. Our current docker mixnet is running several services. List them all, please.
> 
> [DWR] What is a complete list of these services? And which are running in Docker?

Changes applied; so additional questions above.

jdormansteele:
> the node types are:
> 
> mix node service node gateway node
> 
> we no longer have "providers" and also we never had "service providers".

So I guess "provide1" and "provider2" are no longer good names. What should they be?

david415:
> > the node types are:
> > mix node service node gateway node
> > we no longer have "providers" and also we never had "service providers".
> 
> So I guess "provide1" and "provider2" are no longer good names. What should they be?

Gateway node 1
Gateway node 2
Service node 1
Service node 2

david415:
> Production of the PDF is a one-way process based on an ANT transform of the original XML. I included the PDF just as a clarification in case the markdown is confusing. My hope is that the markdown has a high ratio of content-to-markup, unlike XML or HTML. You can't really do a transform roundtrip because the transforms are all lossy. I have to manually transfer the edits back into the XML. I plan to simplify this on my end by diffing the draft markdown with the revised-draft markdown. Then I'll apply revisions as needed, delete old commits, add new ones if needed, and generate a new markdown draft.

we need a way to make the PDF ourselves... and this should be documented either in a Makefile or bash script or something. no?

jdormansteele:
> > Production of the PDF is a one-way process based on an ANT transform of the original XML. I included the PDF just as a clarification in case the markdown is confusing. My hope is that the markdown has a high ratio of content-to-markup, unlike XML or HTML. You can't really do a transform roundtrip because the transforms are all lossy. I have to manually transfer the edits back into the XML. I plan to simplify this on my end by diffing the draft markdown with the revised-draft markdown. Then I'll apply revisions as needed, delete old commits, add new ones if needed, and generate a new markdown draft.
> 
> we need a way to make the PDF ourselves... and this should be documented either in a Makefile or bash script or something. no?

The magical PDF-with-changes-and-comments transform appears to be Oxygen-proprietary. On the other hand, Pandoc, which is GPL2+, does regular PDFs just fine, and is scriptable. You can pass comments of certain source formats to Pandoc, but I haven't experimented with that (yet). I think that round-trips are simply impractical, but one-way transforms are trivial. 

Related to this is Leif's suggestion that just doing everything in the original DocBook code might be the best solution ... one source of truth, changes trackable in GitHub. From that perspective, PDF is nice, but it's not part of the trackable workflow.

PR #682, opened 2024-08-01
Rename docker makefile targets for clarity

### LINKED ISSUE #680: Correct and clarify Docker make target names
In the Docker section, the make targets **make run** and **make start** are named confusingly. This has led to the incorrect reversal of their definitions in https://github.com/katzenpost/katzenpost/blob/main/docker/Makefile. The functionality of the two targets, as shown by experiment, is in fact the following:

**run**: Start the mixnet and run in the foreground
**start**: Start the mixnet, then exit and run in the background

Rather than simply correcting the reversal, these targets could be made self-documenting if renamed as follows:

**run-foreground **: Start the mixnet and run in the foreground
**run-background**: Start the mixnet, then exit and run in the background

david415:
fixes #680


jdormansteele (approved):
The renamings are correct and I have tested the code locally. 

Whoops, I see that Leif is getting rid of the choice altogether. I like the result.

PR #685, opened 2024-08-07
fix 673 sigint

### LINKED ISSUE #673: client2 tests cannot be interrupted with control-c
The shutdown code path of client2 is a little bit messed up since it does not allow us to control-C it to shutdown... when specifically it should do this. This bug only is present when decoy traffic is turned on.

mixmasala:
This PR resolves #673 and fixes some bugs in client2 shutdown path

- **DEBUG: add pprof http listener**
- **client2.NewListener: return err from net.ListenTCP**
- **call EventSink() before SendMessage**
- **client2: client_docker_test: shutdown cleanly on os.Interrupt**
- **client2/client_docker_test: do not listen on prometheus default listening port**
- **client2/Makefile: do not run go test in subshell**
- **do not remotely resolve missing images**
- **client2/Makefile: ensure target image exists**
- **client2/Makefile: allocate tty to pass sigint to container**
- **Revert "DEBUG: add pprof http listener"**

### DIFF
--- a/catshadow/client_docker_test.go
+++ b/catshadow/client_docker_test.go
@@ -846,7 +846,10 @@ loop4a:
 	err = a.RenameContact("b", "b2")
 	require.NoError(err)
 
+	// XXX: a.conversations["b"] - panics with bad map state
+	a.conversationsMutex.Lock()
 	c := a.conversations["b"]
+	a.conversationsMutex.Unlock()
 	require.Equal(len(c), 0)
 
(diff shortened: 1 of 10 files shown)


david415 (approved):
hell yeah, thanks for the bug fixes!

david415:
I think we can just merge this whenever you are ready, @mixmasala We can always open more branches for other bug fixes...

amitbasuri (approved):

PR #686, opened 2024-08-07
dockerdockertest: various improvements

mixmasala:
This refactors the make "dockerdockertest" target in memspool, client, and catshadow packages.

- catch sigint and exit
- verify docker image exists
- remove subshell, call go test directly


david415 (approved):
looks good to me but i did not review thoroughly and i do not understand the docker stuff as well as others

PR #688, opened 2024-08-13
fix client2 dockerdockertest

mixmasala:
this collection of commits fixes issues found while debugging the client2 dockerdockertest make target.

in particular there were several issues with blocked routines at shutdown, a missing "+" required for addressing kaetzchen endpoints, misbehaving fetchInterval loop and aggressive garbage collection timeout resulted in SURB Replies being dropped.

- **Reapply "DEBUG: add pprof http listener"**
- **add -noMixDecoy to genconfig**
- **genconfig: disable client2 decoy traffic with -noDecoy**
- **client2: fix testtest endpoint**
- **client2: connection: run peer reader with worker.Go**
- **client2: listener: eliminate redundant Wait()**
- **client2: Shutdown conn and pki worker before halting client worker**
- **client2: incoming_conn, connection: start workers with worker.Go()**
- **client2/daemon: shutdown listener first**
- **client2/client_docker_test: fix other endpoints**
- **client2/exponential: do not block opCh at shutdown**
- **client2/connection, client2/daemon: update fetchDelay from pki Document**
- **client2/listener: fix comment**
- **client2/client_docker_test: remove invalid constraint**

### DIFF
--- a/client2/client_docker_test.go
+++ b/client2/client_docker_test.go
@@ -147,10 +147,10 @@ func testDockerMultiplexClients(t *testing.T) {
 	message1 := []byte("hello alice, this is bob.")
 	nodeIdKey := hash.Sum256(pingTargets[0].IdentityKey)
 
-	reply := sendAndWait(t, thin1, message1, &nodeIdKey, []byte("echo"))
+	reply := sendAndWait(t, thin1, message1, &nodeIdKey, []byte("+echo"))
 	require.Equal(t, message1, reply[:len(message1)])
 
-	reply = sendAndWait(t, thin2, message1, &nodeIdKey, []byte("echo"))
+	reply = sendAndWait(t, thin2, message1, &nodeIdKey, []byte("+echo"))
 	require.Equal(t, message1, reply[:len(message1)])
 
 	err = thin1.Close()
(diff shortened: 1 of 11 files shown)


client2/exponential.go
```diff
@@ -45,20 +45,20 @@ func (e *ExpDist) OutCh() <-chan struct{} {
...
 }
 
-func (e *ExpDist) Stop() {
-	e.Halt()
-}
-
 func (e *ExpDist) UpdateRate(averageRate uint64, maxDelay uint64) {
-	e.opCh <- opExpNewRate{
+	select {
+	case <-e.HaltCh():
```
david415, line 50: nice catch

client2/client_docker_test.go
```diff
@@ -292,4 +295,9 @@ func testDockerClientSendReceive(t *testing.T) {
 
 func init() {
 	shutdownCh = make(chan interface{})
+	go func() {
+		http.ListenAndServe("localhost:4242", nil)
+	}()
+	runtime.SetMutexProfileFraction(1)
```
david415, line 301: do we want to turn on profiling for the client2 all the time?

client2/exponential.go
```diff
@@ -45,20 +45,20 @@ func (e *ExpDist) OutCh() <-chan struct{} {
...
 		maxDelay:    maxDelay,
+	}:
 	}
 }
 
 func (e *ExpDist) UpdateConnectionStatus(isConnected bool) {
-	e.opCh <- opConnStatusChanged{
-		isConnected: isConnected,
+	select {
+	case <-e.HaltCh():
```
david415, line 60: another good catch

david415 (approved):
lots of really good changes and bug fixes!

mixmasala:
I added a few more fixes; making worker routines be managed by worker.Worker.Go and fixed a few endpoints and enabled the other tests, which should now pass :-)
I eliminated the extra sync.WaitGroup by having the connection workers routines by started by the listener, so when the listener is halted all connection workers should return before listner.Halt() returns.

client2/thin/thin.go
```diff
@@ -280,6 +280,8 @@ func (t *ThinClient) worker() {
 		if err != nil {
 			t.log.Errorf("thin client ReceiveMessage failed: %v", err)
 			if err == io.EOF {
+				// XXX: should we halt ThinClient on EOF??
```
david415, line 283: yes i think we should halt the thin client on EOF... and we should remove that comment and ideally not put them in our code

client2/timer_queue.go
```diff
@@ -64,9 +64,12 @@ func (t *TimerQueue) Len() int {
 }
 
 func (t *TimerQueue) Push(priority uint64, value interface{}) {
-	t.pushCh <- &pushedItem{
+	select {
```
david415, line 67: very nice fix

client2/pki.go
```diff
@@ -207,14 +207,14 @@ func (p *pki) worker() {
 
 func (p *pki) updateDocument(epoch uint64) error {
 	pkiCtx, cancelFn := context.WithCancel(context.Background())
-	go func() {
+	p.Go(func() {
```
david415, line 210: cool

PR #697, opened 2024-08-19
core/log: hold exclusive writer mutex when calling Backend._backend.Log

mixmasala:
Multiple writers shouldn't be able to call Log concurrently

Similarly, SetLevel should also be changed with an exclusive mutex.


mixmasala:
Log() should be thread safe

PR #699, opened 2024-08-19
client/client_docker_test: do not use testify from multiple goroutines

mixmasala:
there are other bugs like this too


david415:
It looks like your changes remove the goroutines from these tests and uses waitForEvent which blocks on the next event... essentially this makes the tests run the operations is a fixed sequence of operations and events. That is fine.

You could have removed testify from the testing goroutines and it might've fixed that problem with testify races... but removing the goroutines probably fixes a bunch of other bugs ;-) anyway... that's my review. it looks good to me but I'm curious how much slower are the tests now with this change? @mixmasala

david415 (approved):
i left my review in the other comment.

PR #723, opened 2024-09-13
Add rough draft reunion spec

david415:


ioerror:
We should reopen when we're going to commit the full spec.

PR #766, opened 2024-10-20
client2: make EventSink() return new channel with all Events

mixmasala:
This makes it possible for BlockingSendMessage to receive and filter interleaved Events
from concurrent calls to BlockingSendMessage.


david415 (approved):
lgtm

PR #781, opened 2024-11-26
core/epochtime: replace WarpedEpoch build flag with KATZENPOST_EPOCH_DURATION environment variable

### LINKED ISSUE #744: Set Epoch duration from env var
Set Epoch duration from env var

david415:
fixes https://github.com/katzenpost/katzenpost/issues/744


leif (approved):

david415:
> also confirmed (to my surprise) that it actually even can work with a _one second_ epoch.

did the resulting consensus document actually contain any mix descriptors or was is an empty doc. i think it was empty but i don't know that's why i'm asking.

PR #782, opened 2024-11-26
client2/thin: fixup client2

david415:
Minor improvements to the client2 thin client's select statements such that it selects on additional events that it previously was not so that it responds to context done and halting events.

Make some correction to client2 logging. Reduce the superfluous verbosity. Correction.


ioerror:
As I said in the call, I think this is fine to merge, sorry to delay on making the comment.

PR #785, opened 2024-11-29
Fix Signing bug + genconfig: default to dilithium

david415:
Here we fix a signing bug that prevented us from using some of the signature schemes made available in hpqc, namely it lets us use the Signature schemes from our fork `circl` library if we've specifically made them available.

### DIFF
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -97,7 +97,7 @@ var (
 	)
 )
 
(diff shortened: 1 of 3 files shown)


ioerror (approved):
Hash-based signatures are a very safe bet. Is there a reason to change the default to Dilithium?

Would this make any practical difference in any deployments such as namenlos? My read is that it would not, can you confirm?

Issue #788, opened 2024-12-02
fix wire protocol DoS attack


david415:
Our Noise protocol messages can be used in a DoS attack if the max message size is made very large. How to avoid that?

within Katzenpost the `MaxMsgLen` is the max size of any PQ Noise message we will send and it is used by our wire session methods, `SendCommand` and `RecvCommand`:

Wtihin SendCommand it's used like this:

```golang
	if ctLen > constants.MaxMsgLen {
		return errMsgSize(ctLen)
	}
```
Likewise in RecvCommand:

```golang
	ctLen := binary.BigEndian.Uint32(ctHdr[:])
	if ctLen < macLen || ctLen > constants.MaxMsgLen {
		return nil, errMsgSize(int(ctLen))
	}

	// Read and decrypt the Ciphertext.
	ct := make([]byte, ctLen)
	if _, err := io.ReadFull(s.conn, ct); err != nil {
		return nil, err
	}
```

within Nyquist, the `HandshakeState` contains a field `maxMessageSize`:

```golang
func (hs *HandshakeState) WriteMessage(dst, payload []byte) ([]byte, error) {
```

The `WriteMessage` uses the `maxMessageSize` like this:

```golang
	if hs.maxMessageSize > 0 && len(dst)-baseLen > hs.maxMessageSize {
		hs.status.Err = ErrMessageSize
		return nil, hs.status.Err
	}
```
`ReadMessage` also uses it:

```golang
func (hs *HandshakeState) ReadMessage(dst, payload []byte) ([]byte, error) {
	if hs.status.Err != nil {
		return nil, hs.status.Err
	}

	if hs.maxMessageSize > 0 && len(payload) > hs.maxMessageSize {
		hs.status.Err = ErrMessageSize
		return nil, hs.status.Err
	}
```

CipherState's `EncryptWithAd` method also uses `maxMessageSize`:

```golang
	if cs.maxMessageSize > 0 && len(plaintext)+cs.aeadOverhead > cs.maxMessageSize {
		return nil, ErrMessageSize
	}
```

Likewise `DecryptWithAd`

```golang
	if cs.maxMessageSize > 0 && len(ciphertext) > cs.maxMessageSize {
		return nil, ErrMessageSize
	}
```

PR #789, opened 2024-12-02
core/wire: get rid of constant max message size

david415:
Here we instead base our max message bounds on the max padding size for a given command set.
Fixes https://github.com/katzenpost/katzenpost/issues/788

### DIFF
--- a/authority/voting/client/client.go
+++ b/authority/voting/client/client.go
@@ -246,7 +246,11 @@ func (p *connector) fetchConsensus(auth *config.Authority, ctx context.Context, 
 		return nil, err
 	}
 	p.log.Debugf("sending getConsensus to %s", auth.Identifier)
-	cmd := &commands.GetConsensus{Epoch: epoch}
+	cmd := &commands.GetConsensus{
+		Epoch:              epoch,
+		Cmds:               commands.NewPKICommands(p.cfg.PKISignatureScheme),
+		MixnetTransmission: false, // Disable padding for direct dirauth transmission
+	}
 	resp, err := p.roundTrip(conn.session, cmd)
 	if err != nil {
 		r, ok := resp.(*commands.Consensus)
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -97,7 +97,7 @@ var (
 	)
 )
 
-var testSignatureScheme = signSchemes.ByName("Ed448-Dilithium3")
+var testSignatureScheme = signSchemes.ByName("ed25519")
 
 func getClientCfg(f string) *config.Config {
 	cfg, err := config.LoadFile(f)
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -97,7 +97,7 @@ var (
 	)
 )
 
-var testSignatureScheme = signSchemes.ByName("Ed25519")
+var testSignatureScheme = signSchemes.ByName("Ed448-Dilithium3")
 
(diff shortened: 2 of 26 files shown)


authority/voting/client/client.go
```diff
@@ -246,7 +246,11 @@ func (p *connector) fetchConsensus(auth *config.Authority, ctx context.Context,
 		return nil, err
 	}
 	p.log.Debugf("sending getConsensus to %s", auth.Identifier)
-	cmd := &commands.GetConsensus{Epoch: epoch}
+	cmd := &commands.GetConsensus{
+		Epoch:              epoch,
+		Cmds:               commands.NewPKICommands(p.cfg.PKISignatureScheme),
+		MixnetTransmission: false, // Disable padding for direct dirauth transmission
```
ioerror, line 252: Is there a security analysis (read: any at all) that says that we want a special case for padding? Wouldn't it just be safer to chunk it up with padding (if over some limit) or pad it (if under that limit)? Having this be distinguishable seems risky - I think the result is that padding is disabled which implies a distinguisher. Is that correct? Can a third party adversary tell that a client is doing a fetch of a specific thing here?

bench/bench_test.go
```diff
@@ -97,7 +97,7 @@ var (
 	)
 )
 
-var testSignatureScheme = signSchemes.ByName("Ed448-Dilithium3")
+var testSignatureScheme = signSchemes.ByName("ed25519")
```
ioerror, line 100: As a wise person once said: ¿Por qué no los dos?

core/wire/commands/old_client_commands.go
```diff
@@ -138,7 +162,11 @@ func (c *MessageEmpty) ToBytes() []byte {
 	binary.BigEndian.PutUint32(out[2:6], uint32(c.Cmds.messageEmptyLength()))
 	out[6] = byte(messageTypeEmpty)
 	binary.BigEndian.PutUint32(out[8:12], c.Sequence)
-	return padToMaxCommandSize(out, c.Cmds.maxMessageLen(c))
+	return c.Cmds.padToMaxCommandSize(out, false)
+}
+
+func (c *MessageEmpty) Length() int {
+	return cmdOverhead + 4 + c.Cmds.geo.PacketLength
```
ioerror, line 169: Why is 4 special and if it is special, can it be given a constant variable name that ensures no one asks this question again?

ioerror (changes requested):

bench/bench_test.go
david415: it just needs to match whatever signature scheme we've selected by default with genconfig and used for the docker mixnet.

authority/voting/client/client.go
david415: We cannot pad interactions with the PKI. Easier to talk on the phone to explain

core/wire/commands/old_client_commands.go
david415: 4 bytes in a uint32

bench/bench_test.go
ioerror: Okay, I'd make it hybird pq by default.

core/wire/commands/old_client_commands.go
ioerror: Okay, understood.

ioerror (approved):

ioerror (approved):

PR #950, opened 2025-12-14
Retry early and often

ioerror:
WIP

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ test-unit: test-config
 # Run replica unit tests (requires RocksDB dependencies)
 test-replica: test-config
 	@echo "Running replica unit tests..."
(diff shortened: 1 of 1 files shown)


authority/voting/client/client.go
```diff
@@ -152,69 +159,56 @@ type connection struct {
...
 		u, err := url.Parse(peer.Addresses[idx])
 		if err != nil {
 			lastErr = fmt.Errorf("%s: invalid URL %s: %v", peerInfo(), peer.Addresses[idx], err)
 			continue
 		}
+
 		ictx, cancelFn := context.WithCancel(ctx)
 		conn, err = common.DialURL(u, ictx, dialFn)
-		defer cancelFn()
+		cancelFn()
```
david415: wrong. should be defer cancelFn()

PR #998, opened 2026-05-06
Merge pr 995 onto main

david415:
## Summary

Merges Jacob Appelbaum's namenlos stabilisation work (PR #995) onto current main, then layers cleanups and a fix on top:

- The merge itself, with conflict resolutions documented commit-by-commit.
- A `DocumentVersion` bump from `v0` to `v1` to reflect the new `LambdaR`, `ConfiguredReplicaIdentityKeys`, and `ReplicaEnvelopeKeys` fields on the PKI document.
- A rename of the dirauth and client config wrapper type `KEMPublicKeyPEM` to `LinkPublicKey`, since the wrapper exists to give the link-key field a `MarshalText` method now that hpqc no longer provides one on the bare KEM interface.
- A correctness fix in `replica/pki.go` so that descriptor uploads target `currentEpoch + 1` rather than `currentEpoch`, removing a latent ordering hazard that was masked on long-epoch deployments and surfaced on the docker mixnet's two-minute epochs.
- A new `CONFIG_CHANGES.md` at the repository root documenting every TOML config change between v0.0.71 and current main, per component, with breaking changes called out for operators upgrading.

## Commits

| Commit | Description |
|---|---|
| `8441f707` | Merge branch 'v0.0.71-bugfixes-systematic' (PR #995) |
| `662ebf37` | core/pki: bump DocumentVersion to v1 |
| `59cf027e` | config: rename KEMPublicKeyPEM wrapper to LinkPublicKey |
| `62e55e6e` | replica/pki: target currentEpoch+1 in publishDescriptorIfNeeded |
| `755bb423` | Add CONFIG_CHANGES.md documenting v0.0.71 to main TOML changes |

## The merge

The branch and main met cleanly at `bc061025`. The conflict surface was 10 files, 31 regions. Notable resolutions:

- `authority/voting/server/state.go`: kept main's full `Document` literal (Lambda* and replica-IDs fields) and folded in the branch's `Version: pki.DocumentVersion` initialisation.
- `client/connection.go`: kept main's `ErrorCode` short-circuit on `Consensus2` reception; folded in branch's improved log messages and chunk-stream error handling.
- `replica/pki.go`: accepted branch's switch-based publish-epoch refactor (`c1a63a72`).
- `replica/outgoing_conn.go`: accepted main's `egressWorker` plus `sendAndRecv` architecture; it subsumes the branch's inline send-and-receive and adds retry-queue handling, transient-error classification, and proxy-manager routing.
- `courier/server/outgoing_conn.go`: kept main's metric and halt-aware infrastructure, took branch's debug logs minus their redundant `DEBUG:` prefix, dropped the log line that referenced the removed `IsRead` field on `ReplicaMessageReply`.
- `authority/voting/client/client_test.go`: kept main's slow-CI timing thresholds and preserved the branch's six new tests covering descriptor-fanout quorum behaviour.
- `docker/Dockerfile.alpine-base` and `docker/Makefile`: accepted main (Alpine 3.21 plus RocksDB 10.2.1, pull-or-build recipe).

Two merge artefacts surfaced under `go vet` and were folded into the merge commit: a duplicate `fetchResult` type in `authority/voting/client/client.go` (both sides had independently introduced the identical type), and a missing `authconfig.KEMPublicKeyPEM{...}` wrapper at `replica/pkiworker_test.go:192`.

## DocumentVersion bump

The PKI document format gained `LambdaR`, `LambdaRMaxDelay`, `ConfiguredReplicaIdentityKeys`, and `ReplicaEnvelopeKeys` after v0.0.71. A v0 receiver cannot parse a v1 document; the version check in `ParseDocument` will refuse it cleanly rather than deserialise it incorrectly.

## LinkPublicKey rename

The wrapper type was named after its serialisation idiom rather than its purpose. It exists solely to give the dirauth and client config link-key fields a `MarshalText` so that BurntSushi/toml can write them; hpqc removed the text marshaller from the bare `kem.PublicKey` interface, so we now depend on this shim for TOML round-tripping. Naming it after the field it wraps reads better at the call site:

```go
vConfig.LinkPublicKey{PublicKey: linkKey}
```

The two unkeyed-literal vet warnings in `core/genconfig/genconfig.go` are converted to keyed at the same time.

## The publish-epoch fix

A descriptor for the document being voted on must claim `Epoch = currentEpoch + 1`, because voting for `doc[N]` happens during epoch `N-1`. The mix server has always done this directly:

```go
// server/internal/pki/pki.go
doPublishEpoch := currentEpoch + 1
```

The replica's `publishDescriptorIfNeeded` did not. Both v0.0.71's original code and Jacob's `c1a63a72` refactor used a two-step pattern that targets `currentEpoch` on the first attempt and only flips to `currentEpoch + 1` on a follow-up tick within the same epoch:

```go
if p.lastPublishedEpoch == 0 {
    doPublishEpoch = epoch                     // first try, will be late
} else if elapsed < PublishDeadline {
    doPublishEpoch = epoch
    if p.lastPublishedEpoch == epoch {
        doPublishEpoch = epoch + 1             // recovery on retry
    }
}
```

The dirauth handler at `onReplicaDescriptorUpload` rejects with `Late descriptor upload` when the document for the descriptor's `Epoch` field has already been finalised. Voting for `doc[X]` runs during epoch `X-1`, so by the time the replica is in epoch `X`, `doc[X]` is already signed. A descriptor with `Epoch: currentEpoch` therefore loses every first attempt; a second attempt within the same epoch hits the recovery branch and succeeds.

This pattern silently relied on the PKI worker's tick interval being short relative to the epoch length. Production deployments such as namenlos use long (hour-scale) epochs, so the worker fires the loop several times per epoch and the recovery path is reached. The docker mixnet runs `epoch_duration=2m` by default; the loop fires once per epoch and never reaches the recovery branch, so every replica's descriptor is rejected `Late`, every consensus document has empty `StorageReplicas`, and pigeonhole writes hit `GetShards: no shard replicas are currently online` from the courier.

The fix replaces both branches with the single line the mix server uses, removing the dependence on a second tick:

```go
doPublishEpoch := currentEpoch + 1
```

`uploadDeadline := PublishDeadline - descriptorUploadSafety` and the budget check are kept as a "skip this epoch and try the next" gate; everything below the decision (envelope keys, descriptor build, `PostReplica`, `lastPublishedEpoch` accounting) is unchanged.

## Config-changes documentation

`CONFIG_CHANGES.md` at the repository root catalogues every TOML config file change between v0.0.71 and current main, organised by component: directory authority, mix server / gateway / service node, replica, courier, client daemon, and thin client. Each section lists added, removed, and renamed fields, with breaking changes called out so operators upgrading from v0.0.71 know what to edit before redeploying. A closing note covers the PKI document format bump from v0 to v1 and the common-defaults timeout adjustments. Field names and example tables were cross-checked against the configs the docker mixnet actually generates and consumes.

## Local verification

The chain was tested end to end against a fresh local docker mixnet:

1. Cold-started the mixnet on `59cf027e` (the merge plus the version bump plus the LinkPublicKey rename, without the publish-epoch fix). Confirmed the failure mode: every replica's descriptor was rejected as `Late descriptor upload`, every consensus document had `0 replica descriptors`, and `make dockertest_new_pigeonhole_alice_sends_bob` failed at `EncryptWrite` with `Internal error` (kpclientd logged `failed to create envelope: GetShards: no shard replicas are currently online`).
2. Applied the publish-epoch fix as `62e55e6e` and ran `make replica-restart`. The new binary's gitcommit appears in the replica logs as `62e55e6e`.
3. After one epoch transition the dirauth logged `Generating document with 7 mix descriptors and 5 replica descriptors`, then `Achieved threshold consensus`. The next round's submissions were `Successfully submitted replica descriptor for epoch ...` during `accept_desc` with ample slack remaining.
4. Re-ran `make dockertest_new_pigeonhole_alice_sends_bob` against the same daemon and dirauths, only the replica binary swapped. The test passed in 46.79s; Bob successfully decrypted Alice's plaintext.

## Test plan

- [ ] Linux unit tests pass.
- [ ] Darwin unit tests (amd64 and arm64) pass.
- [ ] Windows MSYS2 unit tests pass.
- [ ] Replica unit tests pass.
- [ ] Docker mixnet integration tests pass.
- [ ] Spot-check the docker mixnet locally: `make stop start && make wait && make run-ping && make dockertest_new_pigeonhole_alice_sends_bob`.


ioerror (approved):
This may or may not be fine, it includes most of the changes tested on the network but it includes other changes, and those are untested on the namenlos network.

PR #1000, opened 2026-05-06
cmd/fetch: add --min-replicas readiness flag

david415:
A freshly started docker mixnet may seal its first consensus document before any storage replica has had time to upload its descriptor, leaving the document with zero replicas. Any client that consults that document for a pigeonhole operation then fails at GetShards with "no shard replicas are currently online".

Add a --min-replicas (-r) flag that, when non-zero, makes fetch keep consuming NewDocumentEvent messages until a consensus document arrives with at least the requested number of replica descriptors. The default of 0 preserves the existing behaviour of returning the first document received.

This lets CI replace its blanket sleep with a deterministic readiness check, e.g. fetch -r 2 before running any pigeonhole test against a fresh mixnet.

### DIFF
--- a/cmd/fetch/main.go
+++ b/cmd/fetch/main.go
@@ -27,12 +27,14 @@ import (
 	"github.com/katzenpost/katzenpost/client/config"
 	"github.com/katzenpost/katzenpost/client/thin"
 	"github.com/katzenpost/katzenpost/common"
+	cpki "github.com/katzenpost/katzenpost/core/pki"
 )
 
 // Config holds the command line configuration
 type Config struct {
-	ConfigFile string
-	LogLevel   string
+	ConfigFile  string
+	LogLevel    string
+	MinReplicas int
 }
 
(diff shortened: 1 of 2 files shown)


cmd/fetch/main.go
```diff
@@ -120,12 +122,26 @@ func runFetch(cfg Config) error {
...
 		case <-client.HaltCh():
 			return fmt.Errorf("connection closed before receiving PKI document")
 		}
 	}
 }
+
+// hasEnoughReplicas reports whether the given document satisfies the caller's
+// minimum replica requirement. A min of 0 imposes no constraint.
+func hasEnoughReplicas(doc *cpki.Document, min int) bool {
+	if min <= 0 {
```
ioerror, line 143: This would also make sense as an argument, config option, or another method rather than a magic number of '0' that is hard coded.

ioerror (approved):
I would not hard code a 0 but I don't think it matters enough to block merging.

PR #1049, opened 2026-07-03
cmd/dirauth: reproducible build (-trimpath, stripped symbols)

gabrix73:
Plain 'go build' embeds absolute source paths in debug info / build ID, producing different binaries depending on the checkout path. Verified non-reproducibility by building from two different absolute paths and comparing sha256.

Add -trimpath and -ldflags "-s -w" to the dirauth Makefile target. Verified bit-for-bit identical output across different build paths, including with dirauth's CGO post-quantum crypto dependencies (highctidh, falcon, sqisign, sphincsplus/ref) in the import tree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ server:
 	cd cmd/server; go build
 
 dirauth:
-	cd cmd/dirauth; go build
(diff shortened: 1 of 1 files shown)


david415:
Thank you Gab!

PR #1081, opened 2026-08-09
server: optionally persist mix keys on clean shutdown

tirewapped:
Mix keys were generated fresh per boot and never persisted, so a clean restart (make stop/start, an upgrade) left the new instance's private keys mismatched with the consensus the dirauths retained: clients failed the first-hop MAC check at the gateway until a fresh consensus was voted in one epoch later.

This was done for forward secrecy reasons, at the cost of causing up to an epoch of packet loss.

When the new [Server] PersistMixKeysOnShutdown option is enabled, mixkeys.Halt() (clean shutdown only) writes every live per-epoch keypair to DataDir/mixkeys/mixkey-<epoch>.bin atomically, and mixkeys.Generate() reloads them on the next boot. A crash never reaches Halt, so hard failures still rotate the keys (one-epoch outage as before). genconfig enables the option for every docker testnet node; the default stays off to preserve fresh-per-boot forward secrecy in production for now.

Files are keyed by epoch with a magic+kind header (nike vs kem), so a corrupt or scheme-mismatched file fails loudly and falls back to generating a fresh key. Test covers the Persist/Load/Remove roundtrip.

(cherry picked from commit bc47bd73826ffa05c807a3a640c47d5e48d0d5e9)

### DIFF
--- a/cmd/genconfig/main.go
+++ b/cmd/genconfig/main.go
@@ -170,6 +170,11 @@ performance optimization and security requirements.`,
 			"before reaping it (0 means use the daemon's built-in default of 10 minutes; "+
 			"docker mixnet sets a shorter value so the reaper visibly exercises)")
 
+	cmd.Flags().StringVar(&cfg.PersistMixKeysOnShutdownDir, "persistMixKeysOnShutdownDir", "",
+		"enable mix key persistence across clean restarts, writing every live mix key on "+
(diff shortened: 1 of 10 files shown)


tirewapped:
If used in production, this should actually write the keys to tmpfs rather than disk. (this is in progress now)

We'll keep the option off by default.

I think we should make the config option for #1086 mutually exclusive with this. Only one or the other (or neither) should happen on a clean shutdown.

PR #1087, opened 2026-08-10
server: withdraw from consensus before shutdown

### LINKED ISSUE #1086: feat: signal to stop advertising descriptor, to fall out of consensus before shutdown/upgrade
Currently, when a node needs to be upgraded, it degrades the network for the rest of the epoch (and often for the following epoch also).

To allow for nodes to be upgraded or decommissioned, we should have a mechanism for a node to cleanly withdraw itself from the consensus: that is, to continue serving traffic but stop advertising itself, so that in subsequent epochs it will not receive any more traffic and can be shut down.

When this option is enabled, operators using systemd will need to set `TimeoutStopSec` to tell `systemctl stop` to wait long enough for the current epoch to end (eg something >20 minutes with our current default epoch length). The option will be mutually-exclusive with the #1081 option; the tradeoff is of a mix having a brief period of unplanned packet loss vs. being out of service for an entire epoch (but gracefully shifting its load to other servers in the same layer during that time).

gabrix73:
## Summary

- add `[Server] WaitForConsensusExitOnShutdown`, disabled by default
- stop future descriptor uploads and cancel/join an upload already in flight
- keep serving traffic and fetching PKI documents until the node is absent from consensus
- keep fatal-error shutdown immediate
- reject configurations that also enable `PersistMixKeysOnShutdown`

## Shutdown timing

The server advertises for the next epoch. A descriptor upload can also return an aggregate error after enough individual authorities have accepted it. The shutdown path therefore waits through the last epoch that was cached or potentially advertised, rather than assuming the current epoch is sufficient.

With the default 20-minute epoch, the conservative wait can approach 40 minutes. The configuration documentation recommends a systemd `TimeoutStopSec` above that, for example `45min`.

## Testing

- `go test ./server/config ./server/internal/pki`
- `go test ./server -run '^(TestConsensusExitWait|TestShutdownGracefullyHonorsConsensusWithdrawalOption)$'`
- `go test -race ./server/internal/pki`
- `go test -run '^$' ./cmd/server ./server/...`
- `go vet ./server/config ./server/internal/glue ./server/internal/pki`

This is stacked on #1081 so the two shutdown modes can be validated together. It should be rebased onto `main` after #1081 merges.

Closes #1086.

### DIFF
--- a/CONFIG_CHANGES.md
+++ b/CONFIG_CHANGES.md
@@ -138,6 +138,20 @@ None.
 
 Source: `server/config/config.go`.
 
+### `[Server]`
+
+- **Added** `WaitForConsensusExitOnShutdown` (bool, default `false`). When
+  enabled, SIGINT or SIGTERM stops new descriptor uploads while the node
+  continues serving traffic and fetching PKI documents. The daemon exits
+  after the last epoch in which it may still appear has ended. The wait is
+  conservative: an upload attempt counts even if it returned an error,
+  because enough directory authorities may already have accepted it. If the
+  next epoch's descriptor was uploaded before the signal, shutdown can take
+  almost two complete epochs. With the default 20-minute epoch, configure a
+  systemd `TimeoutStopSec` comfortably above 40 minutes, for example `45min`.
+  This option and `PersistMixKeysOnShutdown` are mutually exclusive; enabling
+  both is a configuration error.
+
 ### `[Server.Gateway]`
 
 - **Removed** `[Gateway.UserDB]` table (and its `[Gateway.UserDB.Bolt]`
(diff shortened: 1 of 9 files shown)


tirewapped:
@gabrix73 i re-ran the CI because i think the failure on the first run was an existing intermittent failure unrelated to this branch. (edit: it was; it did pass when re-run.)

gabrix73:
@david415 @tirewapped, PR #1081 has merged. I rebased this branch onto current main and updated the PR base. Local focused tests and race tests pass. The PR is now ready for review; GitHub CI is rerunning on the rebased commit.

tirewapped:
there are some minor fixes here https://github.com/gabrix73/katzenpost/pull/1

other than that this looks good to me, thanks for doing it!

tirewapped:
@gabrix73 i assume you won't object to these changes so i just opened this https://github.com/katzenpost/katzenpost/pull/1092 so @david415 can merge it.

PR #1091, opened 2026-08-14
Replica proxy congestion

david415:
implements changes to address replica/proxy congestion behavior


gabrix73 (changes requested):
The attempt timeout is calculated from the sweep deadline before calling `proxyToShardWithSlot`, but acquiring `proxySema` is not bounded by that timeout. Under saturation, a request can wait indefinitely for a worker slot. Once acquired, the original timeout is still passed downstream, and `sendProxyRequestSync` starts its timer only after MKEM encapsulation and dispatch.

The total wall-clock duration can therefore become semaphore wait plus MKEM work plus the full attempt timeout, exceeding `ProxyRequestTimeout` and violating the documented whole-sweep budget.

Please use the absolute sweep deadline while acquiring the semaphore and recompute the remaining duration before waiting for the reply. A regression test should pre-fill `proxySema` and verify that the attempt returns when its budget expires.

david415:
broke it up into 7 PRs

tirewapped:
note that this was broken up in to #1097 #1098 #1099 #1100 #1101 #1102 #1103

PR #1094, opened 2026-08-15
Add katzenpost_node_ready gauge and make fetch --require-ready poll it

tirewapped:


tirewapped:
part of #1095 which could land on its own

PR #1096, opened 2026-08-15
Add katzenpost_node_ready gauge and make fetch --require-ready poll it

tirewapped:
This adds readiness gauges so the fetch command can block until each component is actually ready, hopefully eliminating some CI flakiness. It also removes some unnecessary delays which were making it take longer than necessary for the first consensus to reach the fetch command.


tirewapped:
cherry-picked from #1095; first opened as #1094 but with wrong branch name.

PR #1099, opened 2026-08-25
replica: document the proxy path as self-tuning (7 of 7)

david415:
Add a "Self-tuning" section to the replica README explaining that the
replica measures its own host at startup, caches the result, and
re-measures when NumCPU or hostname changes. The three config fields
(ProxyWorkerCount, ProxyRequestTimeout, IncomingQueueSize) exist for
research and chaos workloads, not for operators; setting them replaces
a measurement with a guess.

Add a "Is this replica keeping up?" section with a metrics table and
three diagnostic scenarios for proxy_active_attempts vs
proxy_sem_waiters. Fleet-wide saturation is a LambdaR provisioning
matter for the directory authorities, not a per-node knob.

Reword the replicationReplyRetry constant doc to describe
classification semantics rather than prescribing retry, since
handleReplicaWriteReply now counts the drop instead.


tirewapped (dismissed):

tirewapped:
i changed the base to #1103's branch so we can see the actual diff here.

after 1103 lands the base should be switched back to main before this is merged.

tirewapped (approved):

PR #1103, opened 2026-08-25
replica: size the proxy sweep to one ProxyRequestTimeout (6 of 7)

david415:
ProxyWorkerCount defaults to runtime.NumCPU and one attempt can run to
ProxyRequestTimeout, so a slot held across every candidate let NumCPU
stalled requests stop the replica proxying to anything at all, healthy
holders included. That is how one sick shard holder took its peer down
with it on windfall2.

Take the slot per attempt instead, and give the whole sweep a single
budget divided among the candidates still to try, so a client waits
one ProxyRequestTimeout rather than one per holder. The budget covers
the wait for a worker slot, the MKEM encapsulation, and the network
round-trip; each is measured against the same deadline so the total
never exceeds ProxyRequestTimeout regardless of how time distributes
between them.

Do not start an attempt whose share is smaller than one saturated MKEM
operation (measured by the startup self-check): starting it would take
a worker slot and hold a core through a CTIDH1024 keygen and group
action purely to reach a certain timeout.

The wait for the slot is bounded by the sweep deadline so a saturated
pool cannot hold a request past the budget.

Thanks to @gabrix73 for flagging on https://github.com/katzenpost/katzenpost/pull/1091
that the semaphore wait sat outside the sweep budget, and for
suggesting the acquire be bounded by the deadline with a regression
test.


tirewapped (dismissed):

tirewapped:
Note: i changed the base from main to #1102's branch so we can see the correct diff here.

After 1102 lands, the base of this PR should be changed back to main to land this.

replica/integration_tests/proxy_failover_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+	replicaEpoch, _, _ := replicaCommon.ReplicaNow()
+	payload := []byte("failover must degrade to the surviving holder")
+	_, sharding := writeBoxToShards(t, env, writer, payload, replicaEpoch)
+
+	boxID, err := reader.NextBoxID()
+	require.NoError(t, err)
+
+	// Take one holder out. The other still has the box.
+	downed := sharding.ReplicaIndices[0]
```
gabrix73, line 222: `TestProxyFailsOverWhenHolderIsDown` shuts down `ReplicaIndices[0]`, but the production sweep still chooses its first holder with `CryptoRandIndex`. If the live holder is chosen first, this test passes without exercising failover. Could the test force the downed holder to be first (for example by driving the sweep with a fixed order or injecting the selection) so the regression is deterministic?

gabrix73 (commented):
Two non-blocking review observations.

replica/handlers.go
```diff
@@ -919,17 +1036,17 @@ func (c *incomingConn) proxyWriteRequest(replicaWrite *pigeonhole.ReplicaWrite,
...
 	responseCh := c.l.server.proxyManager.RegisterProxyRequest(envelopeHash, mkemPrivateKey, targetEnvelopeKey, replicaMessage, *idHash, targetShard.Name)
 
 	// Dispatch the command to the target replica
+	start := time.Now()
 	c.l.server.connector.DispatchCommand(replicaMessage, idHash)
 	c.log.Debugf("Dispatched proxy request to %s, waiting for response", targetShard.Name)
 
-	// Wait for the response with configurable timeout
-	timeout := time.Duration(c.l.server.cfg.ProxyRequestTimeout) * time.Second
+	// Wait for the response within this attempt's share of the sweep budget.
```
gabrix73: On this timeout path, the waiter returns but the request remains in `pendingRequests` until a late reply or periodic cleanup using the full `requestTimeout`. Now that each attempt gets only a share of the sweep budget, timed-out attempts can remain counted and retained while the sweep has already moved to another holder. Would it make sense to call `FailRequest(envelopeHash)` before returning and add a test that the pending entry is cleared immediately?

gabrix73 (approved):
Both observations from my previous review are addressed. No blocking findings from me. Approved.

PR #1109, opened 2026-08-29
Add Docker release interop tests

ioerror:
Boots HEAD together with the recent releases in one docker testnet and swaps individual services, including a mixed authority quorum, to other versions, checking that consensus still forms and a client still round trips. The interop testnet runs at a larger production topology (6 authorities, 8 mixes, 5 gateways, 4 service nodes, 4 replicas, 3 layers) scoped so the dev testnet is unchanged; consensus needs a simple majority (4 of 6), so a 3/3 version split passes only if both versions interoperate.


david415:
A version-compatibility matrix for the docker mixnet: build three revisions (previous/current/candidate, defaulting to the two newest vX.Y.Z tags plus HEAD), stand up a full network from one revision's own genconfig, then swap each role to each of the other two and re-check consensus, echo ping, HTTP round trip and a pigeonhole write/read. Nine roles x three baselines = 27 CI legs; each leg internally loops all three component versions. The idea is sound and the missing coverage is real. The implementation has some problems.

Blocking

1. The pre-push hook makes the repo effectively unpushable. .githooks/pre-push rejects any pushed commit whose changed files contain a byte outside tab plus printable ASCII. On main today, 77 files already violate that, 76 of them .go, because the SPDX headers use ©:

client/client.go:1: // SPDX-FileCopyrightText: © 2023 David Stainton
authority/voting/server/server.go:80: // ... (paper §3, main.tex:381) ...

Once someone runs make install-git-hooks, touching client/client.go blocks git push until the copyright symbol is stripped. This is also the origin of the unrelated de-unicoding of docker/README.rst in this PR (every — to -, every § to "Section"). A repo-wide ASCII policy deserves its own PR and its own argument; it does not belong in a docker-testing change, and as written it cannot be complied with.

2. Per-service version isolation is timing, not filesystem. All services of a role share one binary path. interop-swap-* and interop-test-quorums do, per service: compose stop -> install/mv onto the shared $(net_name)/replica.alpine -> compose start. Correct assignment depends on the container having exec'd the ELF before the next mv lands. Nothing verifies it afterwards.

This matters most in interop-test-quorums, the one leg whose entire purpose is a heterogeneous quorum (replica2/4 on one version, replica3/5 on the other). If the race loses, the leg silently tests a homogeneous quorum, passes, and runtime-artifacts.tsv reports the mix that was intended rather than the one that ran. Two mitigations are already present by luck: genconfig emits restart: "no" for every service, so nothing picks up a stale binary on a surprise restart, and the next iteration's compose stop widens the window. Also worth noting server.alpine is shared across the mix, gateway and servicenode roles, so "role isolation" is bookkeeping only.

The fix is cheap and already half-built: the ldflags bake versioninfo.Revision=$commit into every binary, so a post-swap assertion comparing the running process's revision against runtime-artifacts.tsv would turn a silent false pass into a hard failure.

3. Hardcoded service lists break every topology knob. interop_services_replica=replica1 ... replica5, interop_services_dirauth=auth1 auth2 auth3, interop_services_mix=mix1 mix2 mix3 are literals, while the network is generated from mixes=3 auths=3 serviceNodes=3 storageNodes=5. make interop-test storageNodes=3 generates replica1-3 and then tries docker compose stop replica4. interop-test-quorums additionally hard-requires at least 3 auths and 5 replicas. These should derive from the topology variables.

CI workflow

4. docker-interop.yml is dead but still fires. on: [pull_request] with both jobs if: ${{ false }} adds a permanently-skipped check to every PR in the repo. on: workflow_dispatch costs nothing and keeps the checks list clean until the suite is ready.

5. The build-to-test artifact handoff likely doesn't work. actions/upload-artifact documents that file permissions are not preserved. interop-build's cache-validity check requires [ -x "$out/$bin.$distro.$artifact" ] for every binary; if the exec bit is lost on download, all 27 test jobs invalidate the cache and rebuild all three revisions from source, defeating the shared build job entirely. Latent, since the jobs are disabled, but it means the workflow has never been proven end to end. A chmod +x after download, or tarring before upload, would settle it.

6. previous_ref: v0.0.98 / current_ref: v0.0.99 hardcoded in the workflow env duplicates the Makefile's tag-derived default and will rot at the next release. The Makefile already picks the two newest stable tags on its own.

Base image

7. A second, unrelated base image with undeclared dependencies. Everything else in docker/ builds katzenpost-$(distro)_base from docker.io/alpine:3.21, which explicitly installs bash perl curl netcat-openbsd gcc musl-dev git coreutils. Dockerfile.interop.alpine instead layers on ghcr.io/katzenpost/alpine-base:rocksdb-10.2.1, an image not built by this repo, and adds only Go. But the interop recipes need bash ($(sh)), perl (interop-http-server.pl) and curl (the metrics readiness probe), none of which are installed or asserted. And the RocksDB choice looks stale: go.sum has zero rocksdb entries and the regular docker path builds replica.alpine on the plain alpine base, so the suite pulls a heavyweight external image for a dependency the code no longer has, while missing the packages it does. Dockerfile.rust sets the precedent with --build-arg BASE_IMAGE=katzenpost-$(distro)_base.

8. The image is pinned by mutable tag, and the stamp hashes the tag string. interop-image-check fingerprints '$(interop_base_image)' + GO_VERSION + Dockerfile. If rocksdb-10.2.1 is re-pushed upstream, the local stamp still matches and you keep building against the cached old image. Pin by digest.

9. Two Go-toolchain installers with different security properties. Dockerfile.base.alpine pipes curl | tar unverified; Dockerfile.interop.alpine parses go.dev's JSON index with positional awk to extract a sha256 and verifies it. The verification is an improvement, but its trust anchor is the same host as the download, and the awk parse is brittle. One shared implementation would be better than two divergent ones.

Smaller

10. interop-validate is tautological. It asserts that the literal lists in the Makefile have the lengths hardcoded in the same recipe. It cannot fail unless someone edits a list and forgets the count, and it breaks outright if interop_roles is overridden. The README bills it as "checks matrix construction", which oversells it.

11. interop-client-check can silently pass on nothing. It runs -test.run "^TestNewPigeonholeAPIAliceSendsBob$", and go test exits 0 when a -run pattern matches nothing. The test does exist at v0.0.98, v0.0.99 and main, so it is correct today; the day it is renamed, the check becomes a no-op that still reports PASS. Grepping the output for --- PASS: TestNewPigeonholeAPIAliceSendsBob would close it.

12. Missing-artifact failures are confusing. $(shell cat $(interop_dir)/$(X)/artifact 2>/dev/null) appears in interop-wait, interop-http-check, interop-client-check and interop-compose, and several of those targets have no interop-build prerequisite. With no cache the expansion is empty and the recipe runs /interop-fetch/fetch.alpine., giving "no such file" instead of "you have not built the artifacts".

13. Hardcoded timeouts among the parameterized ones. interop-wait uses a literal 720-second consensus deadline and timeout 12m, while interop_readiness_timeout, interop_ping_timeout and interop_leg_timeout are all variables.

14. The perl HTTP server's failure is invisible. interop-http-check backgrounds it via docker compose exec -T ... sh -c 'perl ... &' with no nohup/setsid, and exec returns 0 regardless. On the second call the bind fails, perl dies, and the check cannot distinguish "server running" from "server died" (it happens to work because the first instance is still listening).

15. Cosmetic: client_tests is built without -trimpath while every other binary has it. And each baseline re-runs the pure-baseline check once per role (component == baseline in eight of the legs); the check type differs per role so it isn't fully redundant, but against a 45-minute leg budget it is real wall clock.

ioerror:
Thanks for the review - going through it top to bottom, each fix is its own commit on the branch.

> A version-compatibility matrix for the docker mixnet: build three revisions (previous/current/candidate, defaulting to the two newest vX.Y.Z tags plus HEAD), stand up a full network from one revision's own genconfig, then swap each role to each of the other two and re-check consensus, echo ping, HTTP round trip and a pigeonhole write/read. Nine roles x three baselines = 27 CI legs; each leg internally loops all three component versions. The idea is sound and the missing coverage is real. The implementation has some problems.
>
> Blocking

> 1. The pre-push hook makes the repo effectively unpushable. .githooks/pre-push rejects any pushed commit whose changed files contain a byte outside tab plus printable ASCII. On main today, 77 files already violate that, 76 of them .go, because the SPDX headers use ©:
>
> client/client.go:1: // SPDX-FileCopyrightText: © 2023 David Stainton
> authority/voting/server/server.go:80: // ... (paper §3, main.tex:381) ...
>
> Once someone runs make install-git-hooks, touching client/client.go blocks git push until the copyright symbol is stripped. This is also the origin of the unrelated de-unicoding of docker/README.rst in this PR (every — to -, every § to "Section"). A repo-wide ASCII policy deserves its own PR and its own argument; it does not belong in a docker-testing change, and as written it cannot be complied with.

You're right, that was my local gate and it should not have been in the tree - I removed it, and restored the README unicode I had needlessly rewritten, in https://github.com/katzenpost/katzenpost/commit/4a56157ce9fce776fe8b0ceaef84978e82479cc7

> 2. Per-service version isolation is timing, not filesystem. All services of a role share one binary path. interop-swap-* and interop-test-quorums do, per service: compose stop -> install/mv onto the shared $(net_name)/replica.alpine -> compose start. Correct assignment depends on the container having exec'd the ELF before the next mv lands. Nothing verifies it afterwards.
>
> This matters most in interop-test-quorums, the one leg whose entire purpose is a heterogeneous quorum (replica2/4 on one version, replica3/5 on the other). If the race loses, the leg silently tests a homogeneous quorum, passes, and runtime-artifacts.tsv reports the mix that was intended rather than the one that ran. Two mitigations are already present by luck: genconfig emits restart: "no" for every service, so nothing picks up a stale binary on a surprise restart, and the next iteration's compose stop widens the window. Also worth noting server.alpine is shared across the mix, gateway and servicenode roles, so "role isolation" is bookkeeping only.
>
> The fix is cheap and already half-built: the ldflags bake versioninfo.Revision=$commit into every binary, so a post-swap assertion comparing the running process's revision against runtime-artifacts.tsv would turn a silent false pass into a hard failure.

Good catch, fixed now. https://github.com/katzenpost/katzenpost/commit/62be7e4690f88bf1871c74400ee489d4d4797d39

> 3. Hardcoded service lists break every topology knob. interop_services_replica=replica1 ... replica5, interop_services_dirauth=auth1 auth2 auth3, interop_services_mix=mix1 mix2 mix3 are literals, while the network is generated from mixes=3 auths=3 serviceNodes=3 storageNodes=5. make interop-test storageNodes=3 generates replica1-3 and then tries docker compose stop replica4. interop-test-quorums additionally hard-requires at least 3 auths and 5 replicas. These should derive from the topology variables.
>
> CI workflow

Agreed, the service lists derive from mixes/auths/serviceNodes/storageNodes now. https://github.com/katzenpost/katzenpost/commit/852c2a4f0e2afcb61d6f8067e211ffd0c7c6d2fc

> 4. docker-interop.yml is dead but still fires. on: [pull_request] with both jobs if: ${{ false }} adds a permanently-skipped check to every PR in the repo. on: workflow_dispatch costs nothing and keeps the checks list clean until the suite is ready.

Fixed, it runs on workflow_dispatch now. https://github.com/katzenpost/katzenpost/commit/460439fe6d2755715e8e7c19edcd411fc614bb24

> 5. The build-to-test artifact handoff likely doesn't work. actions/upload-artifact documents that file permissions are not preserved. interop-build's cache-validity check requires [ -x "$out/$bin.$distro.$artifact" ] for every binary; if the exec bit is lost on download, all 27 test jobs invalidate the cache and rebuild all three revisions from source, defeating the shared build job entirely. Latent, since the jobs are disabled, but it means the workflow has never been proven end to end. A chmod +x after download, or tarring before upload, would settle it.

Right, chmod +x after download. https://github.com/katzenpost/katzenpost/commit/a8861d3708bace7b4ae1e4b14aa7a4e3961b27dc

> 6. previous_ref: v0.0.98 / current_ref: v0.0.99 hardcoded in the workflow env duplicates the Makefile's tag-derived default and will rot at the next release. The Makefile already picks the two newest stable tags on its own.
>
> Base image

Fixed, make picks the two newest tags itself now. https://github.com/katzenpost/katzenpost/commit/32ff7e123972b7276d2f9d5368393cbfa69a5d3b

> 7. A second, unrelated base image with undeclared dependencies. Everything else in docker/ builds katzenpost-$(distro)_base from docker.io/alpine:3.21, which explicitly installs bash perl curl netcat-openbsd gcc musl-dev git coreutils. Dockerfile.interop.alpine instead layers on ghcr.io/katzenpost/alpine-base:rocksdb-10.2.1, an image not built by this repo, and adds only Go. But the interop recipes need bash ($(sh)), perl (interop-http-server.pl) and curl (the metrics readiness probe), none of which are installed or asserted. And the RocksDB choice looks stale: go.sum has zero rocksdb entries and the regular docker path builds replica.alpine on the plain alpine base, so the suite pulls a heavyweight external image for a dependency the code no longer has, while missing the packages it does. Dockerfile.rust sets the precedent with --build-arg BASE_IMAGE=katzenpost-$(distro)_base.

Agreed. The interop image builds FROM the katzenpost base now, which already ships bash, perl, curl and Go, so it has the deps the recipes need. Kept as its own commit. https://github.com/katzenpost/katzenpost/commit/5c80065d278e7d22008c88d701033cca3637150a

> 8. The image is pinned by mutable tag, and the stamp hashes the tag string. interop-image-check fingerprints '$(interop_base_image)' + GO_VERSION + Dockerfile. If rocksdb-10.2.1 is re-pushed upstream, the local stamp still matches and you keep building against the cached old image. Pin by digest.

Same commit - the base is built from this repo, so there is no external mutable tag to re-push under us. https://github.com/katzenpost/katzenpost/commit/5c80065d278e7d22008c88d701033cca3637150a

> 9. Two Go-toolchain installers with different security properties. Dockerfile.base.alpine pipes curl | tar unverified; Dockerfile.interop.alpine parses go.dev's JSON index with positional awk to extract a sha256 and verifies it. The verification is an improvement, but its trust anchor is the same host as the download, and the awk parse is brittle. One shared implementation would be better than two divergent ones.
>
> Smaller

Same commit - the interop image inherits the base's Go toolchain, so the second installer is gone. https://github.com/katzenpost/katzenpost/commit/5c80065d278e7d22008c88d701033cca3637150a

> 10. interop-validate is tautological. It asserts that the literal lists in the Makefile have the lengths hardcoded in the same recipe. It cannot fail unless someone edits a list and forgets the count, and it breaks outright if interop_roles is overridden. The README bills it as "checks matrix construction", which oversells it.

Fair, it was tautological. It checks real preconditions now - known roles, non-empty derived lists, and the quorum leg's 3 auths / 5 replicas. https://github.com/katzenpost/katzenpost/commit/89d1bc16d478a632fab503196f3439fd97150215

> 11. interop-client-check can silently pass on nothing. It runs -test.run "^TestNewPigeonholeAPIAliceSendsBob$", and go test exits 0 when a -run pattern matches nothing. The test does exist at v0.0.98, v0.0.99 and main, so it is correct today; the day it is renamed, the check becomes a no-op that still reports PASS. Grepping the output for --- PASS: TestNewPigeonholeAPIAliceSendsBob would close it.

Good point, it greps for the specific --- PASS line now, so a no-match fails. https://github.com/katzenpost/katzenpost/commit/c32f633bb6293bab0992b786ac831c40dbd12f11

> 12. Missing-artifact failures are confusing. $(shell cat $(interop_dir)/$(X)/artifact 2>/dev/null) appears in interop-wait, interop-http-check, interop-client-check and interop-compose, and several of those targets have no interop-build prerequisite. With no cache the expansion is empty and the recipe runs /interop-fetch/fetch.alpine., giving "no such file" instead of "you have not built the artifacts".

Fixed, those targets say to run make interop-build now instead of the fetch-path error. https://github.com/katzenpost/katzenpost/commit/11b2bfcba7a735edcfdf5238e9c38d81d77fe80e

> 13. Hardcoded timeouts among the parameterized ones. interop-wait uses a literal 720-second consensus deadline and timeout 12m, while interop_readiness_timeout, interop_ping_timeout and interop_leg_timeout are all variables.

Fixed, it is interop_consensus_timeout now. https://github.com/katzenpost/katzenpost/commit/9849927744240b1a7374e5b2d17cc59867e4a69c

> 14. The perl HTTP server's failure is invisible. interop-http-check backgrounds it via docker compose exec -T ... sh -c 'perl ... &' with no nohup/setsid, and exec returns 0 regardless. On the second call the bind fails, perl dies, and the check cannot distinguish "server running" from "server died" (it happens to work because the first instance is still listening).

Right, it was invisible. The server is idempotent now and I added a readiness probe that fails if nothing is listening on 4242. https://github.com/katzenpost/katzenpost/commit/ae0b9bb6e754b2103b8c00e1f69604a2c8686732

> 15. Cosmetic: client_tests is built without -trimpath while every other binary has it. And each baseline re-runs the pure-baseline check once per role (component == baseline in eight of the legs); the check type differs per role so it isn't fully redundant, but against a 45-minute leg budget it is real wall clock.

Fixed the trimpath. The redundant pure-baseline check I would rather tackle separately. https://github.com/katzenpost/katzenpost/commit/1118d7853d4905081bf0e28ca5e07b1198ac0c38

ioerror:
@david415 thanks again for the review. Each point should be addressed in the fixup commits above. Would you approve if it looks good? Glad to iterate otherwise.

david415:
Checked each fixup against the tree at 5c80065d278e7d22008c88d701033cca3637150a. Eleven of the fifteen are properly closed. One fix is broken in a way that would fail every leg it touches, one is half done, and the point that mattered most was answered on a different target.

Closed: 1, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14 and the trimpath half of 15. On 6, `fetch-depth: 0` does fetch tags, so the Makefile's tag derivation works in the workflow. On 13, the literal `timeout 12m` went along with the 720, so that one is complete.

## 2. The swap verification cannot match, and the quorum leg was not touched

First a correction to my own point 2. `interop-activate` and `interop-swap-*` already used per-service binary paths in 86428938f1f878ef604dbbc654274c831d812c75, so the race I described did not apply to them. The shared path exists in exactly one place, `interop-test-quorums`, and there it is not a race but a no-op.

`interop-compose` rewrites every compose command to the suffixed path (docker/Makefile:860-864):

```
command: /interop-alpine/replica.alpine.replica2 -f /interop-alpine/replica2/replica.toml
```

while docker/Makefile:1159-1160 still installs to `/interop-alpine/replica.alpine`, which no container executes. `interop-activate` never creates that plain path either; it materialises only `$bin.$distro.$service` (docker/Makefile:827-838) plus the two probe binaries. So each quorum swap stops the service, writes a file nothing reads, and starts the service back on its baseline binary. `runtime-artifacts.tsv` records the intended mix, the check passes, and six quorum assignments (three baselines by two orderings) exercise a homogeneous quorum. The README already states the invariant this violates: "Runtime copies include the service name so every container has an immutable binary path."

Second, the verification added in 62be7e4690f88bf1871c74400ee489d4d4797d39 greps `docker compose logs <service>` for `" rev <commit>"` (docker/Makefile:1018). The daemons do not write that line to stdout. `core/genconfig/genconfig.go` sets `Logging.File = ServerLogFile` for the mix server (:690), the replica (:572) and the dirauth (:799), and `core/log/log.go:155-162` points the backend at `os.Stdout` only when `File` is empty, otherwise at the file. Only kpclientd gets `File: ""` (genconfig.go:396), which is why `make client-logs` uses `docker compose logs` while every other watch target tails `katzenpost.log`. So for mix, gateway, servicenode, dirauth and replica the grep can never match: 30 attempts at 2 seconds, then `swap verify: ... did not come up running rev ...` and exit 1, on every swap and on every restore. As committed this turns all 24 role legs red rather than hardening them. For courier and http it is also aimed at the wrong process: those swaps stop and start the servicenode container, then look in that container's output for the courier's or proxy_server's commit.

Reading the daemon's log file rather than the container's stdout would work, for instance grepping `$(net_name)/$service/katzenpost.log` after the restart, with the caveat that the file accumulates across restarts so it has to look only at output written after the swap. Whatever the mechanism, `interop-test-quorums` needs the per-service path and the same assertion, or the leg's entire premise stays untested.

## 3. Half done

The `interop_services_*` variables derive from the topology now, but the two places that materialise per-service binaries are still literal: the sed list in `interop-compose` (docker/Makefile:860-864) and the spec list in `interop-activate` (docker/Makefile:827-838), both fixed at mix1-3, gateway1, servicenode1-3, auth1-3, replica1-5. The down direction is now caught by the new floor in `interop-validate`. The up direction is not:

```
$ make interop-validate storageNodes=7 mixes=5
Interop plan valid: 8 roles x 3 baselines + quorums (auths=3 replicas=7)
```

replica6, replica7, mix4 and mix5 then keep the unsuffixed `/interop-alpine/replica.alpine` and `/interop-alpine/server.alpine` commands, which nothing creates, so those containers fail to start.

## Smaller, arising from the fixes

16. `interop-image-check` now takes `$(distro)_base.stamp` as a prerequisite, and the workflow neither exports nor imports the base image or its stamp. All 27 test jobs therefore rebuild `katzenpost-alpine_base` (apk plus the Go tarball) despite importing the interop image they actually need.

17. The interop stamp fingerprint hashes the base image name, `GO_VERSION` and `Dockerfile.interop.alpine`, but not the base's own stamp. Editing `Dockerfile.base.alpine` rebuilds the base and leaves `katzenpost-alpine_interop` pointing at the old layers. Same staleness as the old point 8, now internal rather than external. Folding `alpine_base.stamp` into the fingerprint closes it.

18. Consolidating on one Go installer kept the unverified `curl | tar` and discarded the sha256 verification. Consolidating was the right call; the checksum is worth keeping in the survivor.

19. `Dockerfile.interop.alpine` is now a bare `FROM ${BASE_IMAGE}`, so the interop image is identical to the base. The separate image, its stamp, and the `podman save` / `podman load` pair could all go.

20. Doc drift: docker/README.rst still says the workflow's "jobs are disabled until the interoperability suite is ready to become a required CI check", which is no longer so now that they are dispatch-only, and still describes `interop-validate` as checking "matrix construction".

## Notes

The workflow has two recorded runs, both skipped, so none of this has executed yet. Once 2 and 3 are settled, a manual dispatch before merge would be worth the wall clock; every finding above is the kind that only a real run flushes out.

The pre-push hook is fine now that the ASCII gate is gone. Two things about it remain worth reconsidering, neither blocking: it runs a full `interop-smoke` (three revisions built, a mixnet up, consensus, ping, HTTP and pigeonhole) on any push touching a `.go` file, and `git diff-tree --check` under `set -eu` still aborts a push on a whitespace warning.

docker/Makefile
```diff
@@ -1,4 +1,4 @@
...
...ourier-restart servicenode1-restart replica-restart replica-test config-only watch-replicas watch-auth watch-mixes watch-courier watch-servicenode watch-gateway watch-all-separate start-with-client-metrics print-url-grafana print-url-prometheus pumba-latency pumba-loss pumba-corrupt pumba-pause pumba-status pumba-stop selfcheck-cache-restore thin-client-sync pigeonhole-cp-genkey pigeonhole-cp-roundtrip check-ports interop-build interop-build-previous interop-build-current interop-build-candidate interop-image-check interop-compose interop-selfcheck-cache-restore interop-start interop-wait interop-run-ping interop-http-check interop-check interop-check-role interop-smoke interop-client-check interop-activate interop-record-artifact interop-swap-mix interop-swap-gateway interop-swap-servicenode interop-swap-dirauth interop-swap-replica interop-swap-courier interop-swap-http interop-swap-client2 interop-restore-role interop-noop interop-test interop-report interop-validate interop-test-baseline interop-test-role interop-test-mixed-network interop-test-quorums interop-stop interop-clean interop-clean-dryrun interop-clean-cache interop-clean-results interop-clean-image interop-clean-all
```
david415, line 1: Too many makefile targets!

docker/Makefile
```diff
@@ -43,6 +43,12 @@ help:
 	@echo " test-distros              - bring up and ping the testnet on every DISTRO, leaving them running (serial, or parallel with make -j)"
 	@echo
 
+install-git-hooks:
```
david415, line 46: what does this do?

docker/Makefile
```diff
@@ -731,13 +1221,13 @@ clean-image-%:
...
 stop-all:
-	@for d in $(STAMP_DISTROS); do $(MAKE) distro=$$d stop; done
+	@for d in $(STAMP_DISTROS); do $(MAKE) distro=$$d stop; $(MAKE) distro=$$d interop-stop; done
 
 # Wipe the network directory of every distro ever built (stamp-inferred).
 # Called by `clean` only; clean-local deliberately touches just the current
 # net_name. `git clean` exits 0 even for a pathspec that matches nothing.
 clean-testnets:
-	@for d in $(STAMP_DISTROS); do git clean -fdx mixnet-$$d; done
+	@for d in $(STAMP_DISTROS); do git clean -fdx mixnet-$$d interop-$$d; done
```
david415, line 1310: too many changes to the makefile. too many makefile targets.

david415 (changes requested):

docker/Makefile
ioerror: This adds a git hook that runs the docker-interop test before pushing - it is optional and local. It reduces the need to use CI that may fail.

docker/Makefile
ioerror: Is there a specific target that you think does not need to be added? Each new target fits the shape of the tasks and allows for testing things discretely.

docker/Makefile
ioerror: What specifically is the issue?

This is the level of complexity that is expected for this kind of feature but it does not impact any other docker targets at all. If none of the `interop-*` targets are used by a developer then they should see no impact.

ioerror:
Thanks for the careful re-review. Starting with the blocking ones, all fixed and confirmed by a local run:

Quorum was a no-op: swaps now install to the per-service path so the container runs the swapped binary, plus the same rev assertion. A clean local run produced a genuinely mixed quorum (auth2/replica2,4 vs auth3/replica3,5) and passed both orderings. https://github.com/katzenpost/katzenpost/commit/c04258e2361f5bd11f714d7cddbb709a0124d1c2

Swap verify: you're right that only kpclientd logs to stdout; server, replica, dirauth, courier and proxy_server log to files. Verify now counts " rev <commit>" in the service's own log tree (katzenpost.log, courier.log, proxy.*.log) and requires it to rise after the restart; validated for every role. https://github.com/katzenpost/katzenpost/commit/d1c6d9930da439a6e2fe929ad0efdd4ac7d64bee

Per-service binaries: interop-compose and interop-activate derive from the topology now, so counts above the old fixed mix1-3/replica1-5 work. https://github.com/katzenpost/katzenpost/commit/f59ef7d3a3c6ead6f854677196d43ca8117e379f

ioerror:
The smaller points:

- 16: the workflow exports/imports the base image and its stamp and touches the stamp, so the test jobs stop rebuilding it. https://github.com/katzenpost/katzenpost/commit/7e28a44b0cb0eb3063ae677a9edabed894da5773
- 17 and 19: the interop image was a bare FROM, so it builds from the base directly now and the save/load and interop stamp are gone; the build fingerprint hashes Dockerfile.base.alpine so base edits rebuild the artifacts. interop-clean-image stays as a documented target that removes the obsolete standalone image. https://github.com/katzenpost/katzenpost/commit/bc5fc66c8631d25f6c271ebb75cbb425bdef52bb
- 18: restored the sha256 verification on the Go download. https://github.com/katzenpost/katzenpost/commit/5db3970f8c643facf9371e46944f2dadce3977a2
- 20: README updated. https://github.com/katzenpost/katzenpost/commit/4f4a6bc59e440195ec2111ec80b485e191f2429a

ioerror:
Two things the local run flushed out, as you predicted:

- The rocksdb-era releases (<= v0.0.98) link grocksdb and cannot build in the rocksdb-free base image. The baseline refs now select only non-rocksdb release tags; with just v0.0.99 available the suite runs candidate vs current and skips previous and quorums until a second clean release exists. https://github.com/katzenpost/katzenpost/commit/c9e0ba2ce21b0d2b315c5e81da7f3b7f947f4b63
- Every interop target now has a one-sentence comment, and an extra-targets CI job exercises the targets the matrix does not (validate, report, test-baseline, test-mixed-network, the clean targets). https://github.com/katzenpost/katzenpost/commit/076ff15dc2e447d3daaff05b7c6e932f7d86a458 and https://github.com/katzenpost/katzenpost/commit/4d11207723d7e2b7fd95d1036283176abfcf8364

I will trigger a manual dispatch run now that this is pushed.

ioerror:
@david415 @tirewapped This is (again) cleanly merging with main. I use it daily to validate changes and to ensure that changes to katzenpost work with different versions as deployed on the network. It has _fewer make targets_ than the current docker testnet and it does not impact that testnet docker config at all.

ioerror:
Minimal alternative that folds the same cross-version interop check into the existing docker testnet instead of a parallel harness: #1137.

ioerror:
I am closing this in favor of  #1137 and we could revisit if needed.

PR #1110, opened 2026-08-31
kpclientd: multiple unix listeners, optional D-Bus name, hardening

ioerror:
- Adds an Addresses field to UnixListenConfig so the daemon can bind several unix sockets at once.
- Accepts each address as a filesystem path or, on Linux, an abstract name with a leading '@', expanded through os.ExpandEnv on both the listen and thin-client dial sides.
- Rejects an empty or bare-'@' address rather than autobinding a random abstract socket.
- Fans multiple sockets into one Listener whose Close drains in-flight connections and joins its accept goroutines, while a single address returns the underlying listener unchanged.
- Retries a member socket's temporary accept failure and retires only that socket on a permanent one instead of failing the whole listener.
- Adds a command-line option and a matching optional configuration item to own a session D-Bus name, off by default and defaulting to kpclientd when the flag is given without one.
- Compiles the D-Bus support only on Linux and stubs it elsewhere, so the godbus dependency never blocks a build on another platform.
- Serializes per-connection socket writes so the shutdown broadcast cannot interleave with the writer goroutine and corrupt the framing.
- Guards incomingConn.appID with a mutex so the reader goroutine no longer races the session-resume swap.
- Fixes the WebsocketListener close race and double-close panic with a sync.Once and surfaces bind errors by binding synchronously.
- Halts the connection before joining the daemon workers so shutdown does not stall on a wedged network send.
- Backs off on temporary accept errors instead of busy-looping.
- Propagates the D-Bus name release error instead of discarding it.


cmd/kpclientd/dbus.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/godbus/dbus/v5"
+)
+
+type busOwner interface {
+	RequestName(string, dbus.RequestNameFlags) (dbus.RequestNameReply, error)
+	Close() error
+}
+
+var sessionBus = func() (busOwner, error) {
```
david415: coding style nitpick:

this line is written weird. a module scoped anonymous function assigned to a variable name. What's the point of an anonymous function which could be a closure but isn't a closure and then naming it? My suggestion here is to rewrite it to a normal function definition like so:

```golang
func sessionBus() (busOwner, error) {
...
}
```

client/listener.go
```diff
@@ -142,6 +144,11 @@ func (l *listener) worker() {
 				l.log.Errorf("Critical accept failure: %v", err)
 				return
 			}
+			select {
```
david415, line 147: This change here seems like a shutdown code path bug fix which could be in it's own separate pull request. Also the accept retry delay is a nice touch.

client/transport/websocket.go
```diff
@@ -101,15 +95,17 @@ func (c *WsListenConfig) Listen() (net.Listener, error) {
...
-		go func() {
-			<-listener.done
-			server.Shutdown(context.Background())
-		}()
-		server.ListenAndServe()
+		<-listener.done
+		server.Shutdown(context.Background())
 	}()
+	// run webserver in go-routine
+	go server.Serve(ln)
```
david415: apparently we're running a webserver

david415:
@ioerror can you please explain the how d-bus is used here? does it allow clients to use d-bus to connect to kpclientd as a thinclient? or does d-bus advertize the sockets listened on?

client/thin/transport/unix.go
```diff
@@ -14,5 +17,5 @@ type UnixDialConfig struct {
 
 // Dial opens a unix-domain-socket connection to c.Address.
 func (c *UnixDialConfig) Dial() (net.Conn, error) {
-	return net.Dial("unix", c.Address)
+	return net.Dial("unix", os.ExpandEnv(c.Address))
```
david415: this seems fine but gosh it kind of makes me think about attacks that could take advantage of this shell variable expansion. I'm not against it but was just thinking about it. Is it the case that you have a specific use case for this Address to use shell variables?

client/transport/unix.go
```diff
@@ -9,4 +9,6 @@ type UnixListenConfig struct {
 	// directory must exist and be writable by the daemon; any stale
 	// socket file at the path is the operator's responsibility.
 	Address string `toml:"Address"`
+	// Addresses are additional sockets bound alongside Address.
```
david415, line 17: cool. this change looks good to me.

client/transport/unix_impl.go
```diff
@@ -5,13 +5,124 @@
...
+			closeListeners(listeners)
+			return nil, err
+		}
+		listeners = append(listeners, listener)
 	}
-	return net.ListenUnix("unix", addr)
+	if len(listeners) == 1 {
+		return listeners[0], nil
+	}
+	return newMultiListener(listeners), nil
```
david415: cool

david415:
The hardening half is genuinely good work and most of it is correct. The two new features both have blocking defects. Two of my findings I proved rather than inferred.

Blocking

1. --dbus-name cannot work in either documented form. Two independent bugs that compound:

- defaultDBusName = "kpclientd" is not a valid D-Bus well-known name (the spec requires two or more period-separated elements). Against the live session bus here:
ownBusName("kpclientd")            -> Requested bus name "kpclientd" is not valid
ownBusName("org.katzenpost.Client") -> <nil>
  So the bare flag always aborts startup with failed to own D-Bus name.
- Setting NoOptDefVal means the space-separated form no longer takes a value. Parsing the PR's own help example:
--dbus-name org.katzenpost.Client   -> dbus-name="kpclientd"  leftover positional=[org.katzenpost.Client]
--dbus-name=org.katzenpost.Client   -> dbus-name="org.katzenpost.Client"
--dbus-name                         -> dbus-name="kpclientd"
  The root command has no Args validator, so the intended name is silently discarded.

Net effect: only --dbus-name=... or the config file reaches a usable name, and the documented invocation fails 100% of the time. Suggest defaultDBusName = "org.katzenpost.kpclientd", = in the help example, and cmd.Args = cobra.NoArgs. dbus_test.go only drives a fake bus, so it structurally cannot catch this; a test asserting the default constant is a syntactically valid well-known name would.

2. d.client.haltConnection() can hang shutdown forever. connection.sendPacket builds an unbuffered errCh and abandons it via <-c.HaltCh(). Halting the connection while a send is in flight leaves the wire worker holding a doneFn that blocks forever on errCh <- err, so onWireConn never returns and Halt()'s Wait() never returns. That is the same stall the change is meant to remove, relocated one goroutine over. Demonstrated with a probe test against this branch:

doneFn BLOCKED: the connection worker goroutine can never return, so Halt()/Wait() hangs

errCh := make(chan error, 1) in sendPacket (and GetConsensus, same pattern) makes the probe pass. Separately, haltConnection halts without setting isShutdown, so onWireConn's deferred onConnStatusChange(wireErr) now takes the error branch during shutdown and runs the ctx.doneFn(ErrNotConnected) drain — the same blocking callback, on a path Shutdown() deliberately suppresses. Set isShutdown before halting.

3. Abstract sockets have no access control. kpclientd does no peer-credential check anywhere in client/ — nothing reads SO_PEERCRED. Today the only thing protecting the thin-client socket is filesystem permissions on its path. An @name socket has none: every process in the network namespace can connect and drive the daemon's mixnet identity. This needs at minimum a prominent warning in the config documentation, and ideally a peer-cred check before abstract binding is offered as a supported option.

Worth fixing before merge

4. Member retirement is silent. multiListener.accept closes and retires a socket on a permanent error with no log at all — TestMultiListenerMemberFailureKeepsServing even asserts the error is not surfaced. The daemon then serves fewer sockets than configured and the operator has no way to know. multiListener has no logger; it needs one.

5. Only one address is visible. Addr() returns listeners[0], so Listening on: %v in listener.worker() names one socket out of N.

6. --validate-only misses bad addresses. The new empty/bare-@ rejection lives in Listen(), but ListenConfig.Validate() only checks the subtable discriminator, and that is what --validate-only runs. Moving the check into Validate() makes the flag do its job. Worth noting the rejection itself is a real fix: pre-PR an empty Address bound @6dff3 (verified) — a socket no thin client could ever find.

7. Schema additions unrecorded and undocumented. [Listen.Unix].Addresses and top-level DBusName are client.toml schema additions; CONFIG_CHANGES.md's client daemon section documents [Listen] and should carry both. Neither the @ abstract form nor $VAR expansion is described anywhere but a code comment inside Listen().

8. Env expansion is Go-only. client/thin/transport/unix.go gained os.ExpandEnv on dial; the Rust and Python thin clients parse the same [Dial.Unix] Address and do not expand. A thinclient.toml using $XDG_RUNTIME_DIR/kpclientd.sock works from Go and fails from the other two.

Minor

9. Removing the if l.closed prologue from WebsocketListener.Accept is a small regression: with both cases ready, select picks at random, so a post-Close Accept can return a queued conn instead of ErrClosed. A default:-guarded <-l.done prologue restores determinism without the data race. Separately, Close still leaks up to 100 queued netConns (pre-existing). And now that binding is synchronous, listener.addr = ln.Addr() would make ws://127.0.0.1:0 usable.

10. TestUnixListenerCloseDrainsInFlight calls require.NoError from a non-test goroutine, i.e. FailNow off the test goroutine.

What is right

The appID race is real and the fix is complete: recvRequest runs on the reader goroutine spawned inside worker(), while handleSessionToken runs on worker() itself, so those two genuinely raced. Every other c.appID reader (doUpdateFromPKIDoc, PickNextRequest, broadcastShutdownEvent, registerConn) is already excluded by connsLock, which the write also holds, and onClosedConn is on the writing goroutine. Nothing is left unguarded.

writeMu is correct and cannot deadlock the shutdown broadcast: the critical section is bounded by perClientWriteDeadline, so broadcastShutdownEvent pins connsLock.RLock for at most one deadline per conn. It also incidentally stops two writers clobbering each other's SetWriteDeadline.

The accept backoff, the closeOnce, the synchronous websocket bind, and the partial-bind cleanup in Listen() are all straightforwardly right. multiListener's close/drain protocol holds up: no deadlock between the worker's deferred Close and listener.Shutdown's Close, and net.ErrClosed does implement net.Error with Temporary() == false (checked), so a fully drained multi-listener is still classified as a critical accept failure rather than spinning.

One process note: this is two features plus six unrelated hardening fixes in a single commit. The hardening half stands on its own and would be easier to land, review, and revert separately from the D-Bus and multi-socket work.

cmd/kpclientd/main.go
```diff
@@ -19,10 +19,13 @@ import (
 	"github.com/katzenpost/katzenpost/common/tomlstrict"
 )
 
+const defaultDBusName = "kpclientd"
```
david415: not a valid d-bus name... you should use that other name you picket with the two dots.

cmd/kpclientd/main.go
```diff
@@ -69,6 +75,9 @@ applications to share a single network connection.`,
 	// Operation mode flags
 	cmd.Flags().BoolVar(&cfg.ValidateOnly, "validate-only", false,
 		"load and validate the configuration file, then exit without side effects")
+	cmd.Flags().StringVar(&cfg.DBusName, "dbus-name", "",
+		"own this session D-Bus name (bare flag defaults to kpclientd)")
+	cmd.Flags().Lookup("dbus-name").NoOptDefVal = defaultDBusName
```
david415, line 81: Setting NoOptDefVal means the space-separated form no longer takes a value.... the PR's help message with example commands is therefore now incorrect.

client/client.go
```diff
@@ -49,6 +49,12 @@ type Client struct {
 	DialContextFn func(ctx context.Context, network, address string) (net.Conn, error)
 }
 
+func (c *Client) haltConnection() {
```
david415, line 52: haltConnection() can hang shutdown forever. connection.sendPacket builds an unbuffered errCh and abandons it via <-c.HaltCh(). Halting the connection while a send is in flight leaves the wire worker holding a doneFn that blocks forever on errCh <- err, so onWireConn never returns and Halt()'s Wait() never returns. That is the same stall the change is meant to remove, relocated one goroutine over. Demonstrated with a probe test against this branch:

doneFn BLOCKED: the connection worker goroutine can never return, so Halt()/Wait() hangs

errCh := make(chan error, 1) in sendPacket (and GetConsensus, same pattern) makes the probe pass. Separately, haltConnection halts without setting isShutdown, so onWireConn's deferred onConnStatusChange(wireErr) now takes the error branch during shutdown and runs the ctx.doneFn(ErrNotConnected) drain — the same blocking callback, on a path Shutdown() deliberately suppresses. Set isShutdown before halting.

client/transport/unix_impl.go
```diff
@@ -5,13 +5,124 @@
...
+		}
+	}
+}
+
+func (m *multiListener) Accept() (net.Conn, error) {
+	select {
+	case conn := <-m.accepted:
+		return conn, nil
+	case <-m.done:
+		return nil, net.ErrClosed
```
david415, line 103: log error

client/transport/unix_impl.go
```diff
@@ -5,13 +5,124 @@
...
+	case <-m.drained:
+		return nil, net.ErrClosed
```
david415, line 105: log error

cmd/kpclientd/dbus.go
ioerror: The weird trick was because the CI does not have dbus. We can mock it differently and so it is now changed as requested.

sessionBus is now a normal function with a small connectBus var that defaults to sessionBus purely for the CI test constraints.

client/transport/unix_impl.go
```diff
@@ -5,13 +5,124 @@
...
 // Listen creates a unix-domain-socket listener bound to c.Address.
 func (c *UnixListenConfig) Listen() (Listener, error) {
-	addr, err := net.ResolveUnixAddr("unix", c.Address)
-	if err != nil {
-		return nil, err
+	addresses := append([]string{c.Address}, c.Addresses...)
+	listeners := make([]Listener, 0, len(addresses))
+	for _, address := range addresses {
+		name := os.ExpandEnv(address)
+		if name == "" || name == "@" {
```
david415: move this address filtering/rejection to ListenConfig.Validate()

client/thin/transport/unix.go
david415: So presumably there will be thinclient.toml configs with shell variables in the Address field.
When that happens the rust and python thinclients will be incompatible with those configs.
Are you planning to add shell variable expansion to all the other thin clients?

david415 (changes requested):

client/listener.go
ioerror: It could be, I agree. I needed to fix the (minor) bug to complete this feature and so I would prefer not to open a third PR (katzenqt pull, this pull) to close it.

For the related enhancement, it was part of the fix to make it work for my other katzenqt flatpak pull reliably. The basic idea is that it is a small hardening of the accept loop rather than a shutdown fix: on a temporary accept error (e.g. EMFILE) it now backs off instead of busy spinning, and it selects on HaltCh so it still exits promptly on shutdown. The reason I didn't make another pull is that I mirrored the delay that the new multiListener uses so they have uniform behavior that is correct.

If this is okay, I would prefer it isn't a separate pull. To clarify: are you requesting a separate pull for this minor thing? Would a separate commit in the same branch be a compromise? Or can we just leave it as is now that I added a little more clarity?

ioerror:
> @ioerror can you please explain the how d-bus is used here? does it allow clients to use d-bus to connect to kpclientd as a thinclient? or does d-bus advertize the sockets listened on?

The dbus service is an activation (systemd user service is started by) and presence handle, not the data path. We could also make it a data path but I had the idea to make a different name for the data path... so the thin clients never talk to kpclientd over dbus; data always flows over the unix socket(s). The bus name does three things: RequestName(DoNotQueue) is a single-instance lock and liveness signal; the systemd user unit is Type=dbus with BusName=network.katzenpost.kpclientd so systemd treats the service as started exactly when the daemon owns the name; and the name is D-Bus-activatable, so the first message starts the unit on demand.

This stuff is part of the changes required to make katzenqt work and not only from a git checkout...

katzenqt’s Flatpak GUI gets --talk-name=network.katzenpost.kpclientd plus a read-only bind of the runtime dir, sends a Peer.Ping to wake the daemon, then connects the thin client over $XDG_RUNTIME_DIR/katzenpost/kpclientd.sock. So dbus decides who runs and when it starts; the socket carries the traffic. All of this is in a systems user service that doesn't impact the entire system, the old abstract Unix socket stays as it was and we gain a file system Unix socket so that the flatpak can talk to the same daemon (and start it as needed or fail knowing it couldn't or that it wasn't running). The reason beyond not changing what works today by adding/providing a new socket is that the flatpak doesn't have network access by design. This couldn't work with only an abstract socket unfortunately. This could work if we didn't have a systemd user service but then katzenqt would need to run kpclientd, so the flatpak (container) would need network access.

Please let me know if this makes sense because it's possible that this is too convoluted or my explanation isn't entirely satisfactory. One problem with many pulls is that the context for some changes is complex to understand.

client/transport/websocket.go
ioerror: Right, the websocket transport runs an HTTP server only to do the WebSocket upgrade handshake. I would rather not touch websocket code in this PR, so I filed #1111 for that plus a couple of related Accept/Close rough edges.

client/transport/unix.go
ioerror: Thanks.

client/transport/unix_impl.go
ioerror: Thanks.

cmd/kpclientd/main.go
ioerror: Fixed: default is now network.katzenpost.kpclientd. Please see commit #a3b94a284

cmd/kpclientd/main.go
ioerror: Fixed: added cobra.NoArgs so the space-separated form errors instead of dropping the name, and the help example uses --dbus-name=. Please see commit #a3b94a284

client/client.go
ioerror: Fixed: buffered errCh in sendPacket and GetConsensus, and set isShutdown before halting so onConnStatusChange short-circuits. Please see commit #c58deea88

client/transport/unix_impl.go
ioerror: Done: the multiListener takes a logger now and logs the retirement. Please see commit #7ba277767

client/transport/unix_impl.go
ioerror: Done: same commit logs the retired socket. Please see commit #7ba277767

client/transport/unix_impl.go
ioerror: Done: moved to UnixListenConfig.Validate, called from ListenConfig.Validate, so --validate-only catches it too. Please see commit #79a7ab706

client/thin/transport/unix.go
ioerror: Right, no shell or subprocess is involved, it is Go's os.ExpandEnv on the config string. But your portability point is the deciding one for me: rather than add expansion to the rust and python clients, I dropped it from the Go dial side, so [Dial.Unix] Address is now read literally by all three. The listen side (kpclientd, Go only) still expands, which is what katzenqt's $XDG_RUNTIME_DIR socket relies on. Please see commit #2adc5e5e4

If katzenqt's flatpak stuff should work differently, please look at the katzenqt https://github.com/katzenpost/katzenqt/pull/42 and we can refactor as needed.

ioerror:
@david415 all of your review comments are fixed now. Would you approve if you are satisfied? Just say the word if anything still needs work.

tirewapped:
I like the idea of kpclientd exposing a unix socket and having the katzenqt flatpak only able to access the outside world via that socket. But I'm not sure that I follow where dbus comes in to play here.

I gather that the idea is that kpclientd will be activated by systemd at the time katzenqt (or another dbus caller) starts? Will it stay running when katzenqt (and/or other applications using it) exits?

Is the idea to have this systemd-activatable kpclientd installed via the katzenqt flatpak, or via some other means?

ioerror:
@tirewapped I covered this for david up-thread: https://github.com/katzenpost/katzenpost/pull/1110#issuecomment-5491570627

Short version: D-Bus is activation and liveness only, never the data path; traffic is always the unix socket. The first D-Bus message activates the systemd user unit on demand, and it stays running after katzenqt exits (ordinary user service, no stop-on-idle). It runs on the host, not in the flatpak (which has no network by design); the unit and service file are installed host-side, see https://github.com/katzenpost/katzenqt/pull/42 .

tirewapped:
but i guess my question is: why are we using dbus, and why are we modifying kpclientd to be able to?

could we not have systemd activate kpclientd with socket activation instead of using dbus?

and, if we **are** going to use dbus, does kpclientd actually need explicit support for it?

i'm pretty sure one can make an arbitrary systemd service be dbus-activatable - the service only needs dbus support if the systemd unit is `type=dbus` but if it is `type=exec` i believe it can still be activated by a dbus service which defines `SystemdService=...`..

ioerror:
@tirewapped socket activation starts a process but signals nothing about readiness: connect() succeeds before the daemon can serve, with no clean way to ask "is it up?". Owning the name gives that. With Type=dbus systemd marks the unit started only once the name is owned, so katzenqt's Peer.Ping doubles as a readiness probe, and RequestName(DoNotQueue) is a single instance lock. A Type=exec unit with SystemdService= activates without owning the name, but then the call has no owner to answer, so you learn start was triggered, not that it is ready.

Support is optional and off by default, tested, and it makes the katzenqt flatpak run today. Glad to compare a concrete socket activation design, but none has been proposed here that covers readiness and the flatpak portal.

david415:
> Owning the name gives that.

I don't think we need or want D-Bus just to know when kpclientd is ready. Right now the thinclient library can simply redial until it connects. Perhaps pause for 200ms after each redial iteration. Anyway, the D-Bus thing is solving a problem which we did not expect to have solved since we already have a solution to that.

What I would like instead is a much more minimal pull-request that doesn't have any D-Bus stuff in it at all.

ioerror:
> > Owning the name gives that.
> 
> I don't think we need or want D-Bus just to know when kpclientd is ready. 

We both want and need it in the context of a flatpak that has no network.

>Right now the thinclient library can simply redial until it connects. 

The thinclient cannot simply redial and ever succeed if the kpclientd isn't running.

> Perhaps pause for 200ms after each redial iteration.

That does not start the daemon or handle stale sockets (e.g. if left on disk by an intermediate program) nor does it signal that the thinclient should now check the consensus.

>  Anyway, the D-Bus thing is solving a problem which we did not expect to have solved since we already have a solution to that.
> 

How so?

As far as I understand the situation we have no solution for automatically starting kpclientd and signaling that the daemon is running. In this repo, we do not even have the systemd user service for kpclientd in use - only in katzenqt.

> What I would like instead is a much more minimal pull-request that doesn't have any D-Bus stuff in it at all.

By not having dbus, we still don't solve the core problem raised by a flatpak of katzenqt without network access.

ioerror:
@tirewapped  @david415: 

I split the work into separate PRs so each piece can be reviewed on its own, and the dbus question is isolated to one of them. That should answer the ask for a minimal PR with no dbus in it but no dbus doesn't actually resolve the reason for all of this work in the first place.

Why it's in katzenpost, and why kpclientd is modified: kpclientd lives here, so the socket and dbus capabilities have to exist here for katzenqt to use them. Otherwise I carry a patch or a pinned hash in katzenqt until this merges. Merging changes nothing about how katzenpost itself runs: no dbus usage happens in katzenpost, and the name ownership is off by default. The name, the systemd user service, and the flatpak talk-name all live on the katzenqt/flatpak side.

Why dbus and not just socket activation: dbus isn't the data path, and the point isn't mixnet-routing readiness (connected with consensus so traffic flows), which is usually quick and sometimes impossible anyway. It is about keeping kpclientd up so a thin_client can connect and poll consensus state at all. dbus does two things: activation (a Peer.Ping autostarts kpclientd from its .service file if it isn't running) and liveness plus single-instance (RequestName(DoNotQueue); a crashed daemon no longer owns the name, so its absence is a simple explicit signal that it is not running rather than an implicit one inferred from a socket that could be stale). In practice the daemon isn't reliably up: it crashes, it loops when there's no consensus, and it can crash-loop until the systemd user service's fail limit. Whether the socket is actually accepting, the thin_client confirms by polling it. Socket activation only starts a process; it does not tell you the daemon is alive versus crashed, and it does not recover a crash.

On the Type=exec + SystemdService= point: yes, you can dbus-activate a plain unit without the daemon owning the name, but then nobody really owns the name, so you lose the liveness and single-instance signal that owning it gives, which is the point. It would require a bunch of extra work to figure out who opened the socket, if the socket is in some funky state, if there is a race condition, and so on. There is a big space of confusion that exists with this paradigm.

The PRs are now a choose your own adventure of small changes:

- #1115 fix shutdown hang and connection races. No dbus. The deadlock resolved in #1115 was discovered and confirmed in this pr (#1110). This is a fix that is important to land independent of all others to fix bugs.
- #1116 repair a stale unix socket on bind. Again no dbus. This is the stale-socket-on-disk case: a crashed daemon's leftover socket no longer blocks a restart. This is largely orthogonal to any other changes but even when we fix the bugs that can leave a stale socket, we still see that #1116 is necessary but not sufficient for resolving the full set of issues raised about a socket absent a signal with an owner. The #1116 PR should be merged regardless of anything else as it fixes bugs that we actually encounter regularly.
- #1117 support multiple unix listener sockets. No dbus. Stacked on #1116 which should be merged first. Keeps the abstract socket in use today and adds multiple unix sockets in a single config such that we can have any combination of abstract unix sockets or filesystem unix sockets. This implementation depends on merging #1116.
- #1118 multiple unix listener sockets with stale-socket repair. No dbus. In anticipation that we might want one PR that fixes the stale sockets and allows for multi sockets, I combined #1116 and #1117 into one PR if you would rather review a single change. Merge this instead of those two, or those two instead of this.
- #1119 add optional session dbus name for activation. The only one with dbus, off by default and Linux only. This is where the dbus discussion belongs, and even with #1116 landed, we can see that activation and liveness are needed beyond a file system unix socket's mere presence.

So the minimal no-dbus path is #1115 plus either #1118 or the pair #1116 and #1117. dbus is a separate decision in #1119 and blocks none of the rest. All of them are ready to go and can be merged as soon as they are approved by a reviewer.

The thing I still need pinned down for the socket PRs: are you saying keep the abstract socket and add a filesystem socket, or turn the abstract socket into a filesystem socket? #1117 and #1118 do the former, keeping the abstract one people use today and adding the filesystem one the flatpak needs, per the promise not to break APIs already in use. If you mean the latter, please say so and ideally why.

If after all of the above work still somehow leads us to a place where we can't agree - I would be open to understanding a proposal for a design that still activates a host daemon from a network-less sandbox, handles the crash and stuck cases, keeps existing users working, and adds no network attack surface by default. I would request that we open an issue to discuss it on katzenqt's repo in this case. Ideally if someone wants to take that path then posting a short writeup on an issue would be the best next step but I don't think I understand a coherent proposal well enough to open the issue myself. Regardless, I will commit to immediately review any such proposal.

In the meantime I would like to merge either this pull request or the new smaller pull requests above. Eventually, we may find that the only sticking point is dbus, and that is somehow useful progress even if it took us nearly entire week to arrive here. This is somewhat unfortunate but at least this process uncovered tons of bugs that we can fix even if we can't agree on the dbus aspect.

ioerror:
I am closing this because it no longer cleanly merges after other pull requests have been merged. The primary remaining item that remains open is #1119. We should move our discussion about dbus to that pull request.

PR #1112, opened 2026-09-01
client/transport: harden the websocket listener

### LINKED ISSUE #1111: kpclientd websocket transport: harden Accept/Close, reconsider embedded HTTP server
The websocket thin-client transport (client/transport/websocket.go) has a
few rough edges surfaced during review of #1110. Filing separately so
#1110 stays scoped to the unix-socket and D-Bus work and does not modify
the websocket transport.

1. Embedded HTTP server. WsListenConfig.Listen starts a net/http server
   only to perform the WebSocket upgrade handshake. Worth documenting and
   deciding whether a lighter-weight upgrade path is preferable.

2. Accept is non-deterministic after Close. Accept selects over the
   connections channel and the done channel; with both ready, select
   chooses at random, so a post-Close Accept can return a still-queued
   conn instead of net.ErrClosed. A default-guarded <-done prologue fixes
   it without a data race.

3. Close leaks queued connections. The connections channel is buffered
   (cap 100) and is not drained on Close, so accepted-but-undelivered
   netConns are leaked rather than closed.

4. Ephemeral port not reported. With synchronous binding, setting
   listener.addr from the bound net.Listener would make ws://host:0
   report its real port via Addr().

ioerror:
Fixes #1111.

- sync.Once Close plus a done-guarded Accept prologue remove the
  closed-bool data race and double-close, and make Accept deterministic
  after Close.
- Bind synchronously so a bind failure is returned instead of swallowed.
- Report the real bound address so ws://host:0 is usable.
- Drain queued connections on Close so none are leaked.

@bfix you wrote this transport originally, tagging you for review.

### DIFF
--- a/client/transport/websocket.go
+++ b/client/transport/websocket.go
@@ -73,6 +73,8 @@ func (l *WebsocketListener) Addr() net.Addr {
 type WsListenConfig struct {
 	// Address is the URL of the websocket like "ws://localhost:12345"
 	Address string `toml:"Address"`
+	// InsecureSkipVerify disables Origin checking; dev-only, keep false in production.
+	InsecureSkipVerify bool `toml:"InsecureSkipVerify"`
 }
 
 // Listen creates a websocket listener bound to c.Address.
@@ -111,7 +113,9 @@ func (c *WsListenConfig) Listen() (net.Listener, error) {
 	// start a webserver to handle websocket connections
 	mux := http.NewServeMux()
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		conn, err := websocket.Accept(w, r, nil)
+		conn, err := websocket.Accept(w, r, &websocket.AcceptOptions{
+			InsecureSkipVerify: c.InsecureSkipVerify,
+		})
 		if err != nil {
 			return
 		}
--- a/client/transport/websocket.go
+++ b/client/transport/websocket.go
@@ -18,7 +18,10 @@ type WebsocketListener struct {
 	addr        net.Addr      // address of websocket
 	connections chan net.Conn // incoming connections
 	done        chan struct{} // channel "done" signal
+	ln          net.Listener  // bound TCP listener owned by this listener
+	server      *http.Server  // webserver serving the websocket handshake
 	closeOnce   sync.Once
+	closeErr    error
 }
 
 // Accept incoming websocket connection.
@@ -36,10 +39,17 @@ func (l *WebsocketListener) Accept() (net.Conn, error) {
 	}
 }
 
-// Close websocket listener.
+// Close websocket listener. Safe to call more than once.
 func (l *WebsocketListener) Close() error {
 	l.closeOnce.Do(func() {
 		close(l.done)
+		// free the bound port synchronously, then stop the webserver
+		if l.ln != nil {
+			l.closeErr = l.ln.Close()
+		}
+		if l.server != nil {
+			l.server.Close()
+		}
 		for {
 			select {
 			case conn := <-l.connections:
@@ -49,7 +59,7 @@ func (l *WebsocketListener) Close() error {
 			}
 		}
 	})
-	return nil
+	return l.closeErr
 }
 
 // Addr returns the address of the websocket.
@@ -116,13 +126,10 @@ func (c *WsListenConfig) Listen() (net.Listener, error) {
 		return nil, err
 	}
 	listener.addr = ln.Addr()
-	server := &http.Server{Handler: mux}
-	go func() {
-		<-listener.done
-		server.Shutdown(context.Background())
-	}()
+	listener.ln = ln
+	listener.server = &http.Server{Handler: mux}
 	// run webserver in go-routine
-	go server.Serve(ln)
+	go listener.server.Serve(ln)
 
 	// return listener instance
 	return listener, nil
--- a/client/transport/websocket.go
+++ b/client/transport/websocket.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"sync"
 
 	"github.com/coder/websocket"
 )
@@ -17,13 +18,15 @@ type WebsocketListener struct {
 	addr        net.Addr      // address of websocket
 	connections chan net.Conn // incoming connections
 	done        chan struct{} // channel "done" signal
-	closed      bool          // listener closed?
+	closeOnce   sync.Once
 }
 
 // Accept incoming websocket connection.
 func (l *WebsocketListener) Accept() (net.Conn, error) {
-	if l.closed {
+	select {
+	case <-l.done:
 		return nil, net.ErrClosed
+	default:
 	}
 	select {
 	case conn := <-l.connections:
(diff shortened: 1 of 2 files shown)


david415:
The headline is that the PR's central promise is not kept. Close() signals done and hands server.Shutdown to a goroutine, but the bound net.Listener is never closed, so after Close() returns nil the port is still bound and still accepting. That was demonstrated directly: rebinding the address gives bind: address already in use and a fresh dial still succeeds. Since the PR now holds ln in a local at line 114, the fix is very nearly a one-liner that was left undone, and doing it properly (storing ln and server on the struct and closing them under one mutex) would collapse the drain race, the Accept non-determinism and the unbounded Shutdown at the same time.

Two security items deserve attention beyond the diff. Address gets no scheme or host validation, so an omitted field, or a bare localhost:12345 copied from the sibling TCP subtable, binds the whole thin-client control API on 0.0.0.0:80; the PR's new synchronous bind is precisely what makes that listener come up rather than fail quietly. And websocket.Accept still runs with InsecureSkipVerify: true, which disables Origin checking, so any page in the operator's browser can open the loopback socket and drive the thin-client API. That line is untouched by the diff, but it sits inside the function this hardening PR rewrites.

The three new tests do not cover either fix: nothing asserts the real bound port for ws://host:0, nothing runs Accept concurrently with Close (so -race in linux.yml cannot catch a regression of the very race being fixed), and TestWebsocketListenerCloseIdempotent has no assertions at all.

My recommendation would be to ask for changes rather than merge: the direction is right, but the resource ownership needs fixing at the root before this counts as hardening.

bfix:
"@bfix you wrote this transport originally, tagging you for review."
Actually I only copied most of the code over from the TCP listener implementation, so maybe that has the same problems and it would make sense to evaluate changes there too?

ioerror:
i generally agree with the points in the review but some of it seems to miss what I understand as being part of the work by @bfix as I explain inline:

> The headline is that the PR's central promise is not kept. Close() signals done and hands server.Shutdown to a goroutine, but the bound net.Listener is never closed, so after Close() returns nil the port is still bound and still accepting. That was demonstrated directly: rebinding the address gives bind: address already in use and a fresh dial still succeeds. Since the PR now holds ln in a local at line 114, the fix is very nearly a one-liner that was left undone, and doing it properly (storing ln and server on the struct and closing them under one mutex) would collapse the drain race, the Accept non-determinism and the unbounded Shutdown at the same time.

Yes, you are right. The fix stores the net.Listener and http.Server on the struct and closes them under closeOnce, so Close frees the bound port synchronously before it returns, which also folds in the drain race and the Accept non-determinism. https://github.com/katzenpost/katzenpost/commit/1ca7e2fba58e60f18951eb3669b92c81ea1b09f4

> Two security items deserve attention beyond the diff. Address gets no scheme or host validation, so an omitted field, or a bare localhost:12345 copied from the sibling TCP subtable, binds the whole thin-client control API on 0.0.0.0:80; the PR's new synchronous bind is precisely what makes that listener come up rather than fail quietly.

Sane defaults ensure we can test things without guessing. I kept the :80/:443 port default but made a missing host default to loopback instead of the 0.0.0.0 wildcard, so an omitted or bare address binds localhost rather than every interface. If the defaults are wrong, we should pick something different - and @bfix being the primary author, I hope he can drop by to tell us or ack the changes; let's try not to block on this and defer to him. https://github.com/katzenpost/katzenpost/commit/4fb4789863d3f82fb03a852727174a8e2ad11c4a

> And websocket.Accept still runs with InsecureSkipVerify: true, which disables Origin checking, so any page in the operator's browser can open the loopback socket and drive the thin-client API. That line is untouched by the diff, but it sits inside the function this hardening PR rewrites.

InsecureSkipVerify is an opt-in flag that defaults to off, so a real deployment enforces Origin - browser pages are rejected while the thin-client, which sends no Origin header, still connects - and skipping the check is available only for local development when the flag is set on purpose. My understanding (which could be wrong) is that this is how one needs to develop without valid TLS certs (ping @bfix: is this correct?). https://github.com/katzenpost/katzenpost/commit/9f4d8887ec3af785fef907550dcf72ca4cc0e660

> The three new tests do not cover either fix: nothing asserts the real bound port for ws://host:0, nothing runs Accept concurrently with Close (so -race in linux.yml cannot catch a regression of the very race being fixed), and TestWebsocketListenerCloseIdempotent has no assertions at all.

Added: tests for the resolved port and that Close frees it, Accept run concurrently with Close under -race, real assertions in the idempotent test, the loopback default, and Origin enforcement both on by default and off under the flag. https://github.com/katzenpost/katzenpost/commit/16bbc5a85b24b257a507768dba25b291e8e3a7d4

> My recommendation would be to ask for changes rather than merge: the direction is right, but the resource ownership needs fixing at the root before this counts as hardening.

The branch is updated.

ioerror:
Thanks, I checked. The core issues were specific to the websocket wrapper, a custom type around an http.Server and a channel whose Close never closed the listener. TcpListenConfig.Listen returns net.ListenTCP directly, so its Close frees the port and concurrent Accept/Close is handled by the standard library. The only shared thing is the address default, where a hostless ":port" binds 0.0.0.0, and I am happy to set whatever default you prefer.

How would you test this to be sure I did not break it for you? Any advice is welcome. And if it looks right, a review that clicks the approve button would be a big help, unless you have feedback for me to integrate.

ioerror:
@david415 the branch now covers every item from your review. Would you approve if it reads well? Happy to make further changes otherwise.

bfix:
Because I was asked to comment:

(1) if there is a code problem that creates leaking or pending connections please correct them. In a sense I created an sloppy experimental implementation without stating that explicitly enough - definitely my fault. So thanks for looking at it and correcting issues. I have no test cases, so I can only compile code and see how it works.

(2) for the security issues (like no certificate for kpclientd or 0.0.0.0 as default bind address for the listener) my understanding and approach was based on the following:

* kpclientd must run locally as it handles private key material and that should not be done on a remote server. Websocket connections therefore  have to originate from the local machine, but not necessarily from processes running on the same localhost (e.g. when using Docker containers). But I have no problem if a missing host spec is replaced by localhost instead of 0.0.0.0

* I question the usefulness of a kpclientd certificate for websockets: How are users supposed to handle this? If kpclientd ships with a certificate, will that be based on a root CA known to common browsers? What will the certificate certify without being useless or to restricting? How will you prevent anyone from extracting the private key for the certificate from kpclientd? In essence: disabling TLS verification on the websocket was intentional. But of course I am open to good arguments.

david415 (approved):

ioerror:
Thanks for the reply, @bfix. That helps.

Bind address: agreed. A missing host now defaults to loopback rather than `0.0.0.0` ([4fb4789](https://github.com/katzenpost/katzenpost/commit/4fb4789863d3f82fb03a852727174a8e2ad11c4a)). An explicit host is still honored, including `0.0.0.0` for the container case, so this should not affect any legitimate use case, though I could have that wrong; if these changes break something for you, I am happy to help fix it.

Certificate: we agree, and we did not change the certificate requirements. What changed is websocket Origin checking ([623cef4](https://github.com/katzenpost/katzenpost/commit/623cef47d381adab47ec64b489f4b311d42ec4c0)), not TLS. The listener is still plain `ws://`. Origin checking defaults on and can be disabled with `InsecureSkipVerify` ([9f4d888](https://github.com/katzenpost/katzenpost/commit/9f4d8887ec3af785fef907550dcf72ca4cc0e660)) for local or cross-origin dev.

PR #1113, opened 2026-09-02
replica: size the proxy sweep to one ProxyRequestTimeout (6 of 7)

tirewapped:
ProxyWorkerCount defaults to runtime.NumCPU and one attempt can run to
ProxyRequestTimeout, so a slot held across every candidate let NumCPU
stalled requests stop the replica proxying to anything at all, healthy
holders included. That is how one sick shard holder took its peer down
with it on windfall2.

Take the slot per attempt instead, and give the whole sweep a single
budget divided among the candidates still to try, so a client waits
one ProxyRequestTimeout rather than one per holder. The budget covers
the wait for a worker slot, the MKEM encapsulation, and the network
round-trip; each is measured against the same deadline so the total
never exceeds ProxyRequestTimeout regardless of how time distributes
between them.

Do not start an attempt whose share is smaller than one saturated MKEM
operation (measured by the startup self-check): starting it would take
a worker slot and hold a core through a CTIDH1024 keygen and group
action purely to reach a certain timeout.

The wait for the slot is bounded by the sweep deadline so a saturated
pool cannot hold a request past the budget.

Thanks to @gabrix73 for flagging on https://github.com/katzenpost/katzenpost/pull/1091
that the semaphore wait sat outside the sweep budget, and for
suggesting the acquire be bounded by the deadline with a regression
test.


tirewapped:
(this is the same as #1103, which was accidentally merged into #1102's branch instead of main)

PR #1114, opened 2026-09-02
server: instrument seven silent packet drop sites

david415:
These sites discarded a packet with nothing but a Debugf line, four of them on the gateway's terminal delivery path where every echo ends. They now increment the aggregate and per-reason counters like every other drop site, so the reason breakdown accounts for them.

### DIFF
--- a/server/internal/gateway/gateway.go
+++ b/server/internal/gateway/gateway.go
@@ -181,12 +181,16 @@ func (p *gateway) onSURBReply(pkt *packet.Packet, recipient []byte) {
 	geo := p.glue.Config().SphinxGeometry
 	if len(pkt.Payload) != geo.PayloadTagLength+geo.ForwardPayloadLength {
 		p.log.Debugf("Refusing to store mis-sized SURB-Reply: %v (%v)", pkt.ID, len(pkt.Payload))
+		instrument.PacketsDropped()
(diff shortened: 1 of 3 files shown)


ioerror:
@david415  Do you suspect that this is part of or the entire basis of the packet loss issue we have discussed?

david415:
> @david415 Do you suspect that this is part of or the entire basis of the packet loss issue we have discussed?

these changes help us gather statistics about why packets may be dropped. i am in fact not seeing packet loss right now within the mixnet, in any case, such comments are not appropriate or specific to this pull request.

ioerror:
> > @david415 Do you suspect that this is part of or the entire basis of the packet loss issue we have discussed?
> 
> these changes help us gather statistics about why packets may be dropped.

Right, I agree. I see the changes and they generally make sense.

>  i am in fact not seeing packet loss right now within the mixnet,

Interesting, I wonder what changed on namenlos.

> in any case, such comments are not appropriate or specific to this pull request.

Huh, how so?

One strategy for finding the mysterious packet loss would require accounting for each possible drop site and the accounting was missing in these sites for the cases you added. The fix is simple and elegant for increasing accounting but the overall scope is not in the issue's text which is why I asked the question...

ioerror (approved):
LGTM

PR #1116, opened 2026-09-03
client/transport: repair a stale unix socket on bind

ioerror:
Repairs a stale unix socket left behind by a crashed kpclientd.

If kpclientd is killed or crashes, its filesystem socket file survives
and the next start fails to bind with EADDRINUSE. This detects that case:
on EADDRINUSE, if the path is a socket that no live daemon answers on (a
dial is refused), it removes the stale file and rebinds once.

A socket a live daemon still answers on is left in place, so a second
instance cannot evict a running one. Abstract (@ prefix) sockets never
outlive their owner, so EADDRINUSE there is a live owner and is passed
through unchanged. A non-socket file at the path is never removed.

The bind is factored into a small per-address helper so the multi-socket
change can route each of its sockets through the same repair.

Tests: a stale socket is repaired, a live socket is preserved, a regular
file is left untouched, and an abstract address in use is passed through.

The repair applies on POSIX platforms; on Windows the listener keeps its previous behavior.

Relationship: this is one half of the combined unix-listener change. If
the combined PR is merged, close this one. If this and the multi-socket
PR are merged, close the combined PR.

### DIFF
--- a/client/transport/unix.go
+++ b/client/transport/unix.go
@@ -6,7 +6,8 @@ package transport
 // UnixListenConfig configures a unix-domain-socket listener.
 type UnixListenConfig struct {
 	// Address is the path to the unix socket file. The parent
-	// directory must exist and be writable by the daemon; any stale
-	// socket file at the path is the operator's responsibility.
+	// directory must exist and be writable by the daemon; a stale
+	// socket file left by a crashed daemon is cleared on the next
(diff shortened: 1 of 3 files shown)


tirewapped:
superseded by https://github.com/katzenpost/katzenpost/pull/1118 which is now merged.

PR #1117, opened 2026-09-03
kpclientd: support multiple unix listener sockets

ioerror:
Let kpclientd bind more than one unix listener socket.

Based on the stale-socket repair branch (PR base is that branch), so this
PR's own diff is just the multi-socket feature. Merge the stale-socket PR
first.

We already had a single unix socket listener. This adds
[Listen.Unix].Addresses so the daemon can bind several at once, in any
combination of filesystem paths and leading-@ abstract names. A single
Address behaves exactly as before, so existing configs are unaffected.
Addresses are validated up front, and a bind failure on any one closes
the sockets already opened. Each socket is bound through the stale-socket
repair, so a crashed daemon's leftover socket does not block a restart.

Coverage: TestUnixListenerMultiple binds several sockets and accepts a
connection on each; TestUnixListenerAbstractAndFile does the same across
a mixed abstract and filesystem set; and the multiListener tests cover a
member failing, all members failing, and temporary accept errors.

Relationship: this is the other half of the combined unix-listener
change. If the combined PR is merged, close this one and the stale-socket
PR. If this and the stale-socket PR are merged, close the combined PR.

### DIFF
--- a/CONFIG_CHANGES.md
+++ b/CONFIG_CHANGES.md
@@ -338,6 +338,16 @@ The client TOML had the most substantial reshape, driven by the
       Address = "/var/run/katzenpost/kpclientd.sock"
   ```
 
+- **Added** `[Listen.Unix].Addresses` (string array). Optional. The unix
+  listener already bound a single `Address`; with `Addresses` it can bind
+  any number of unix sockets, each a filesystem path or, on Linux, an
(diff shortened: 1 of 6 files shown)


tirewapped:
superseded by #1118 which is now merged.

PR #1118, opened 2026-09-03
kpclientd: multiple unix listener sockets with stale-socket repair

ioerror:
Multiple unix listener sockets, and repair of a stale socket left by a
crashed daemon. This folds the stale-socket repair and multi-socket
support into one change, since the multi-socket bind path uses the repair
for every socket.

[Listen.Unix].Addresses lets the daemon bind several unix sockets at
once, in any combination of filesystem paths and leading-@ abstract
names. A single Address behaves exactly as before, so existing configs
are unaffected. Addresses are validated up front, and a bind failure on
any one closes the sockets already opened.

A crashed kpclientd leaves its filesystem socket file behind, so the next
bind would fail with EADDRINUSE. Each socket is bound through a repair
step: on EADDRINUSE, if the path is a socket that no live daemon answers
on (a dial is refused), the stale file is removed and rebound once. A
socket a live daemon still answers on is left in place, so a second
instance cannot evict a running one. Abstract sockets never outlive their
owner, so EADDRINUSE there is a live owner and is passed through.

Coverage: stale socket repaired, live socket preserved, regular file left
untouched, abstract-in-use passed through; multiple sockets each accept a
connection, mixed abstract and filesystem sets, and multiListener member
failure, all-members failure, and temporary accept errors.

The repair applies on POSIX platforms; on Windows the listener keeps its previous behavior.

Relationship: this combines the stale-socket repair PR and the
multi-socket PR into one. If this is merged, close both of those. If you
prefer to review them discretely, merge those two instead and close this
one.

### DIFF
--- a/CONFIG_CHANGES.md
+++ b/CONFIG_CHANGES.md
@@ -338,6 +338,16 @@ The client TOML had the most substantial reshape, driven by the
       Address = "/var/run/katzenpost/kpclientd.sock"
   ```
 
+- **Added** `[Listen.Unix].Addresses` (string array). Optional. The unix
+  listener already bound a single `Address`; with `Addresses` it can bind
+  any number of unix sockets, each a filesystem path or, on Linux, an
+  abstract-namespace name with a leading `@`, in any combination. A
+  thin_client may connect on any of them. An abstract socket has no
+  filesystem permissions, so any process in the network namespace can
+  connect; access control relies on kpclientd running per user (a systemd
+  user service or run by the user), and a peer-credential check is
+  deferred until there is a system-wide daemon.
+
 - **Added** `PigeonholeGeometry` (table). Pigeonhole protocol
   parameters; required for new pigeonhole channel operations.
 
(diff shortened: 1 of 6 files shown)


client/transport/transport.go
```diff
@@ -22,6 +22,10 @@ type Listener interface {
 	Addr() net.Addr
 }
 
+type Logger interface {
```
david415: why do we need this new Logger interface?

CONFIG_CHANGES.md
```diff
@@ -338,6 +338,16 @@ The client TOML had the most substantial reshape, driven by the
...
 
+- **Added** `[Listen.Unix].Addresses` (string array). Optional. The unix
+  listener already bound a single `Address`; with `Addresses` it can bind
+  any number of unix sockets, each a filesystem path or, on Linux, an
+  abstract-namespace name with a leading `@`, in any combination. A
+  thin_client may connect on any of them. An abstract socket has no
+  filesystem permissions, so any process in the network namespace can
+  connect; access control relies on kpclientd running per user (a systemd
+  user service or run by the user), and a peer-credential check is
+  deferred until there is a system-wide daemon.
```
david415: i don't understand why you are commenting here about a system-wide daemon.

tirewapped (approved):

PR #1119, opened 2026-09-03
kpclientd: add optional session dbus name for activation

ioerror:
Adds an optional session dbus name to kpclientd, off by default and Linux
only, behind a --dbus-name flag and DBusName config. The godbus
dependency is stubbed elsewhere so it never blocks other builds.

Rationale, ordered by how a thin_client uses the name:

1. Presence and activation. Is the daemon there? If not, a Peer.Ping
   autostarts it on the host from its .service file. The katzenqt GUI
   runs in a flatpak with no network and cannot spawn a daemon itself,
   so it needs this.
2. Liveness. Is kpclientd running? The name is held only while the
   process runs, so it is a simple explicit signal of the daemon's
   presence rather than an implicit one inferred from a socket. A crashed
   kpclientd stops owning the name at once.
3. Network. Is a mixnet reachable? A separate signal the thin_client gets
   by talking to kpclientd, e.g. polling consensus. When kpclientd is up,
   katzenqt can show useful state, for example a widget saying the daemon
   works but consensus has not formed yet.
4. Single instance. RequestName(DoNotQueue) keeps exactly one owner.
5. Flatpak security. --talk-name is the least-privilege way for the GUI
   to start and reach a host daemon, with no network or process-spawning
   rights granted to the sandbox.

Companion changes harden this same path: one repairs a stale unix socket
on the next bind, another fixes shutdown hangs and connection races. Even
with all of them landed, socket activation alone cannot report whether
kpclientd is up, and a socket's presence, including one freshly rebound
by the repair, does not mean a live daemon is behind it: it may be down,
crash-looping to its restart limit, or stuck without consensus. The dbus
name is held only by a running owner and drops the instant it dies, and
in a flatpak with no network or spawn rights it is the only way for the
GUI to activate and reach a host daemon.

Relationship: independent of the unix-listener changes; merges in any
order. Only touches CONFIG_CHANGES.md in common with them, a trivial
conflict.

### DIFF
--- a/CONFIG_CHANGES.md
+++ b/CONFIG_CHANGES.md
@@ -338,6 +338,11 @@ The client TOML had the most substantial reshape, driven by the
       Address = "/var/run/katzenpost/kpclientd.sock"
   ```
 
+- **Added** top-level `DBusName` (string). Optional. A session dbus
+  well-known name the daemon owns for its lifetime, for single-instance
+  ownership and dbus activation. Empty (the default) owns no name and
+  needs no bus; the `--dbus-name` flag overrides it.
+
 - **Added** `PigeonholeGeometry` (table). Pigeonhole protocol
   parameters; required for new pigeonhole channel operations.
 
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -261,6 +261,9 @@ type Config struct {
 	// "10m" or "30s".
 	SessionGracePeriod time.Duration
 
+	// DBusName is the session dbus name to own; empty owns none.
+	DBusName string
+
 	upstreamProxy *proxy.Config
 }
 
--- a/cmd/kpclientd/dbus.go
+++ b/cmd/kpclientd/dbus.go
@@ -1,0 +1,37 @@
+//go:build linux
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/godbus/dbus/v5"
+)
+
+type busOwner interface {
+	RequestName(string, dbus.RequestNameFlags) (dbus.RequestNameReply, error)
+	Close() error
+}
+
+func sessionBus() (busOwner, error) {
+	return dbus.ConnectSessionBus()
+}
+
+var connectBus = sessionBus
+
+func ownBusName(name string) (func() error, error) {
+	conn, err := connectBus()
+	if err != nil {
+		return nil, err
+	}
+	reply, err := conn.RequestName(name, dbus.NameFlagDoNotQueue)
+	if err != nil {
+		conn.Close()
+		return nil, err
+	}
+	if reply != dbus.RequestNameReplyPrimaryOwner {
+		conn.Close()
+		return nil, fmt.Errorf("dbus name %s is already owned", name)
+	}
+	return conn.Close, nil
+}
--- a/cmd/kpclientd/dbus_stub.go
+++ b/cmd/kpclientd/dbus_stub.go
@@ -1,0 +1,9 @@
+//go:build !linux
+
+package main
+
+import "errors"
+
+func ownBusName(string) (func() error, error) {
+	return nil, errors.New("dbus name ownership is unsupported on this platform")
+}
--- a/cmd/kpclientd/dbus_test.go
+++ b/cmd/kpclientd/dbus_test.go
@@ -1,0 +1,62 @@
+//go:build linux
+
+package main
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/godbus/dbus/v5"
+)
+
+type fakeBus struct {
+	reply  dbus.RequestNameReply
+	err    error
+	closed bool
+}
+
+func (b *fakeBus) RequestName(string, dbus.RequestNameFlags) (dbus.RequestNameReply, error) {
+	return b.reply, b.err
+}
+
+func (b *fakeBus) Close() error {
+	b.closed = true
+	return nil
+}
+
+func TestOwnBusName(t *testing.T) {
+	original := connectBus
+	t.Cleanup(func() { connectBus = original })
+	want := errors.New("test")
+	connectBus = func() (busOwner, error) { return nil, want }
+	closeBus, err := ownBusName("test")
+	if !errors.Is(err, want) || closeBus != nil {
+		t.Fatalf("unexpected result: close=%t, err=%v", closeBus != nil, err)
+	}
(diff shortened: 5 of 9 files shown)


ioerror:
To any reviewers: please review the discussion on #1110 before requesting that we rehash the basic need for dbus.

david415:
## Automated review

**Bottom line: I would not merge this as-is.** The feature's primary interface doesn't work as documented — verified by actually building and running the binary, not just reading the code.

### Blocking

1. **`cmd/kpclientd/main.go:79`** (interacting with the `Args: cobra.NoArgs` added at line 66) — The conventional space-separated form `--dbus-name custom.name` is silently broken. Because the flag is registered with `NoOptDefVal`, pflag never consumes the following token as its value; that token becomes a stray positional argument, which the new `Args: cobra.NoArgs` then rejects. Reproduced: `kpclientd -c foo.toml --dbus-name custom.name --validate-only` → `ERROR: Unknown command "custom.name" for "kpclientd"`, with the flag silently taking its default value instead. Only `--dbus-name=custom.name` works. This breaks the invocation style used by systemd `ExecStart` lines and shell scripts generally.
2. **`cmd/kpclientd/dbus.go:16`** — `sessionBus()` uses `dbus.ConnectSessionBus()`, which autolaunches a `dbus-launch` subprocess with no timeout when no session bus is discoverable, rather than `dbus.SessionBusPrivateNoAutoStartup()`. Reproduced in a sandbox with no session bus reachable: the autolaunch attempt fires in ~79µs. On a headless/systemd/docker deployment (the exact context this daemon documents running in) with `--dbus-name` set, this either leaks an orphan `dbus-daemon` per restart while the activation feature silently no-ops, or blocks daemon startup indefinitely if the handshake stalls.

### Should fix before merge

3. **`cmd/kpclientd/main.go:115`** — `ownBusName()` runs synchronously *before* the SIGTERM-handling goroutine starts, with no timeout on the connect call. A hang here (see #2) means a `systemctl stop` during startup gets no graceful shutdown and escalates to SIGKILL.
4. **`client/config/config_impl.go:44`** — `DBusName` is never validated anywhere: not in `FixupAndValidate` (unlike every other address/name field in the same struct), not in `--validate-only` (which returns before the dbus block even runs), not in `kpconfig`'s client validator. A malformed name only surfaces as a raw D-Bus error at real startup.
5. **`.github/workflows/linux.yml`** (and darwin/windows/`make test-unit`) — `cmd/kpclientd` is not in any CI test matrix. The new `dbus_test.go`/`main_test.go` are the first tests ever added to this package and never run automatically.
6. **`core/genconfig/genconfig.go`** — No genconfig flag or docker Makefile variable sets `DBusName`, so the real (non-faked) `ConnectSessionBus`/dbus-ownership code path has zero integration coverage — only unit tests against a fake exercise it.

### Worth fixing

7. **`cmd/kpclientd/main.go:110`** — Can't force-disable a config-set `DBusName` via the CLI: `--dbus-name=` (explicit empty) is indistinguishable from the flag being unset, contradicting `CONFIG_CHANGES.md`'s own claim that "the flag overrides it."
8. **`cmd/kpclientd/dbus.go:25`** — A plain "no bus present" connection failure and a genuine "name already owned" single-instance conflict are both wrapped into the same fatal error, aborting startup either way. The former arguably shouldn't block the mixnet client at all.
9. **`go.mod:6`** — The new `godbus/dbus/v5` require line breaks alphabetical order, and `go.sum` gets a byte-for-byte duplicate of a pre-existing `v5.0.4` line plus new entries appended at EOF instead of in sorted position — mechanically confirmed via `go mod edit -fmt`, indicating hand-editing rather than `go mod tidy`.
10. **`cmd/kpclientd/main.go:110`** — The new dbus setup (precedence resolution, ownership, deferred release with its own error logging) is inlined into `runClientDaemon` rather than extracted into a helper, per this project's own function-length style guidance.
11. **`cmd/kpclientd/dbus.go:22`** — `ownBusName`'s `(func() error, error)` return shape is a novel resource-lifecycle idiom with no precedent elsewhere in the codebase, managed via a bespoke `defer` entirely outside the daemon's own `Start`/`Shutdown`/`Wait` lifecycle.

### Minor

12. **`cmd/kpclientd/dbus_stub.go:1`** — The `!linux` stub appears unnecessary; `godbus` compiles cleanly on windows/darwin (verified via local cross-compile), so a single unguarded implementation would both simplify and gain test coverage on other platforms.
13. **`cmd/kpclientd/dbus.go:16`** — The real autolaunch path (including a typed-nil-boxed-into-interface case on `ConnectSessionBus` failure) has zero test coverage; all tests go through a fake `connectBus`.
14. **`cmd/kpclientd/dbus.go:29`** — `conn.Close()` errors are silently discarded in both of `ownBusName`'s error branches.
15. **`cmd/kpclientd/dbus_test.go:36`** — The table-driven test's success path is special-cased by comparing to the literal string `"success"` rather than an explicit `wantErr` field; renaming that case would silently break the assertion.

ioerror:
Thanks @david415, all fifteen of your concerns should now be addressed, one commit each:
1. --dbus-name value works; the bare-flag default is gone, so the flag now needs a value.
2. SessionBusPrivateNoAutoStartup plus Auth and Hello; no dbus-launch is ever spawned.
3. Ownership runs under a signal.NotifyContext with a 10 second timeout.
4. config.ValidateDBusName checks the well-known-name grammar and length in FixupAndValidate and for the flag.
5. cmd/kpclientd is in the linux, darwin, and windows matrices.
6. genconfig --dbusName and a kpclientd_dbus_name Makefile variable write it into the generated client.toml.
7. --dbus-name= (explicit empty) disables a configured name; the flag wins whenever it is set.
8. No reachable bus logs a warning and the daemon carries on; only a name conflict is fatal. Say if you would rather keep that fatal too.
9. go mod edit -fmt, and the go.sum lines are in sorted position with the duplicate dropped.
10. Name resolution and ownership live in ownConfiguredDBusName.
11. ownBusName returns an io.Closer.
12. The non-linux stub and build tags are gone.
13. A test points DBUS_SESSION_BUS_ADDRESS at a missing socket and asserts a fast errNoSessionBus.
14. Close errors are joined into the failure paths.
15. The ownership table has an explicit wantErr.

Assuming that the CI passes, do I need to make any further changes for merge approval?

PR #1120, opened 2026-09-03
Ping probabilistic packet loss attribution

david415:


ioerror (approved):
nice, this is thorough. wire compat looks right (omitempty + the both-directions decode test), and the stats check out to me - wilson/newcombe/bonferroni plus the exclusive-tally trick so a lost packet doesn't get pinned on a node's healthy peer. the sim tests sell it.

one small non-blocking thing: routeNames names hops off CurrentDocument() rather than the doc makePath picked, so at an epoch boundary the names could come from a newer doc than the path. best-effort and nil-safe so it won't break a send - maybe thread the selected doc through?

PR #1123, opened 2026-09-03
client: wait longer than the worst-case round trip in sendAndWait

ioerror:
Fix a CI flake in the docker end-to-end tests (TestDockerMultiplexClients
and other sendAndWait callers).

sendAndWait waited a fixed one minute for a reply, but the docker
testnet's own worst-case round trip is about 54s (the run-ping banner
reports it). A reply that draws the tail of the per-hop Poisson delay,
plus crypto and scheduling overhead the Sphinx delay does not encode and
that -race inflates further, can exceed a minute and time out a healthy
reply, failing all of retrySubtest's attempts.

This raises the wait to three minutes so only a genuinely lost reply
times out. Test-only; retrySubtest still bounds attempts. It is the
fixed-timeout counterpart to the derived timeout cmd/ping already uses.

It does not address the separate 1-hour suite hang seen on some runs,
which is a stalled copy in the courier path.

### DIFF
--- a/client/docker_test_helpers.go
+++ b/client/docker_test_helpers.go
@@ -24,6 +24,10 @@ import (
 const (
 	defaultThinClientConfigFile = "testdata/thinclient.toml"
 	defaultTestLogLevel         = "DEBUG"
+
+	// Above the docker testnet's ~54s worst-case round trip, so a
+	// tail-latency reply is not mistaken for a lost one.
+	replyWaitTimeout = 3 * time.Minute
 )
 
 var (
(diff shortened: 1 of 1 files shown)


david415:
we don't want to want much longer than we need to. this is a bad change in my opinion. instead to make the test less flakey it would be better to wait for the RTT + slop.

ioerror:
@david415  The CI changes all together produce a much more reliable CI and those changes are based on measuring. In isolation this change produced fewer CI failure runs. Previously it was close to a measured optimal outcome but it is the wrong value given the ratio of failures that we see. The result does not wait much longer _overall_ than we would otherwise need to except in a pathological case. Without this change, that pathological case causes us to re-run the entire CI job. This takes much longer and can be up to 60 minutes.

PR #1124, opened 2026-09-03
client/thin: bound the resend wait with an optional context

ioerror:
Give a caller a way to bound how long StartResendingEncryptedMessage
waits for the daemon. The wait had no deadline, so a stalled operation
blocked the caller indefinitely; in the docker end-to-end suite a copy
that never completed hung the whole run until its one-hour cap.

This threads a context through the wait and adds
StartResendingEncryptedMessageWithContext. The three existing methods
pass context.Background(), so reads that wait for a not-yet-written box
keep their unbounded behavior; only a caller that opts into a context can
time the wait out. Underlying copy stalls are fixed at the courier
separately (the copy-replication PR); this is defense in depth.

### DIFF
--- a/client/thin/pigeonhole.go
+++ b/client/thin/pigeonhole.go
@@ -5,6 +5,7 @@ package thin
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


david415 (approved):
looks good to me

PR #1127, opened 2026-09-04
client/config: fix race condition in TestUnmarshalTOML

david415:
The test was mutating shared gateway data across subtests, causing a race condition that manifests as memory corruption on Windows with the race detector enabled. Fix by copying the gateway map before modifying Addresses in each subtest, preventing mutation of the shared parsed TOML data.


ioerror (approved):
As long as we do not mutate the config in the real usage then this test makes sense and is a reasonable change. LGTM.

PR #1132, opened 2026-09-06
cmd: add version flags to remaining commands

### LINKED ISSUE #1131: cmd: add consistent version flags to the four remaining commands
### Summary

Issue [#922](https://github.com/katzenpost/katzenpost/issues/922) standardized version reporting for most commands by using the shared Cobra/Fang path in `common.ExecuteWithFang`.

The current `main` branch still has four exceptions:

- `cmd/courier` supports only the legacy `-v` flag and prints the plain result of `common.Version()`.
- `cmd/echo-plugin` has no version flag.
- `cmd/http-proxy-client` has no version flag.
- `cmd/http-proxy-server` has no version flag.

This follow-up tracks the remaining scope without reopening the completed historical issue.

### Expected behavior

- All four commands accept both `-v` and `--version`.
- Version output is consistent with the styled output used by the other Katzenpost commands.
- A version request exits successfully before configuration loading, logging setup, socket creation, client connections, or other runtime side effects.
- The existing `courier -v` invocation remains compatible.
- Focused tests cover both aliases and successful early exit, following the repository's existing CLI test conventions.

### Implementation notes

- Reuse the shared version and CLI helpers in `common/` where appropriate, preferably the established Cobra/Fang and `common.ExecuteWithFang` pattern.
- Keep the change limited to version handling and avoid unrelated command-line interface changes.
- Take care not to write plugin startup data to stdout when only version information was requested.

### Files in scope

- `cmd/courier/main.go`
- `cmd/echo-plugin/main.go`
- `cmd/http-proxy-client/main.go`
- `cmd/http-proxy-server/main.go`

### References

- Original completed issue: https://github.com/katzenpost/katzenpost/issues/922
- Previous audit comment: https://github.com/katzenpost/katzenpost/issues/922#issuecomment-5373194390

gabrix73:
## Summary

- add the shared Cobra/Fang CLI path to `courier`, `echo-plugin`, `http-proxy-client`, and `http-proxy-server`
- support both `-v` and `--version` with consistent version output
- exit before configuration loading, connections, sockets, or service startup
- preserve the legacy single-dash long flags used by the plugin launcher and existing proxy commands

## Testing

- `go test -count=1 ./cmd/courier ./cmd/echo-plugin ./cmd/http-proxy-client ./cmd/http-proxy-server`
- `go test -run '^$' ./cmd/courier ./cmd/echo-plugin ./cmd/http-proxy-client ./cmd/http-proxy-server`
- `go vet ./cmd/courier ./cmd/echo-plugin ./cmd/http-proxy-client ./cmd/http-proxy-server`
- CLI smoke tests for both version aliases with legacy arguments
- `git diff --check`

Fixes #1131

### DIFF
--- a/cmd/courier/main.go
+++ b/cmd/courier/main.go
@@ -4,9 +4,11 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
 
 	kpcommon "github.com/katzenpost/katzenpost/common"
 	"github.com/katzenpost/katzenpost/common/tomlstrict"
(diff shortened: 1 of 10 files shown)


ioerror (dismissed):
Thanks for hacking on Katzenpost! Minor changes would be most welcome.  I checked that the Kaetzchen launcher (cbor_plugins.go:390) passes -key value as separate tokens so normalizeLegacyArgs rewrites them. A few small notes follow.

cmd/echo-plugin/main.go
```diff
@@ -50,24 +57,49 @@ func (e *Echo) RegisterConsumer(s *cborplugin.Server) {
...
+	cmd.SetArgs(normalizeLegacyArgs(os.Args[1:]))
+	kpcommon.ExecuteWithFang(cmd)
+}
+
+func normalizeLegacyArgs(args []string) []string {
+	normalized := append([]string(nil), args...)
+	for i, arg := range normalized {
+		switch strings.SplitN(arg, "=", 2)[0] {
+		case "-log_dir", "-log_level":
+			normalized[i] = "-" + arg
```
ioerror: This rewrites any token matching a flag name, including a value token, so a config value that happens to equal e.g. -log_dir would get mangled -- please stop at a -- terminator or only rewrite tokens in flag position. Same in all four commands.

cmd/http-proxy-client/main.go
```diff
@@ -137,8 +140,42 @@ func (k *kttp) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
 func main() {
-	flag.Parse()
-	client, err := getThinClient(*cfgFile)
+	cmd := newRootCommand()
+	cmd.SetArgs(normalizeLegacyArgs(os.Args[1:]))
+	kpcommon.ExecuteWithFang(cmd)
+}
+
+func normalizeLegacyArgs(args []string) []string {
```
ioerror: Can we pull this into one common helper like NormalizeLegacyLongFlags(args, names...) instead of copy-pasting it (and its test) four times?

cmd/courier/main_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"bytes"
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/charmbracelet/fang"
+	"github.com/spf13/cobra"
+)
+
+func TestVersionFlags(t *testing.T) {
```
ioerror, line 13: Please add one case that runs a legacy flag through to Run, so a future normalize/flag-name mismatch actually fails a test -- right now only the version short-circuit is covered.

cmd/courier/main.go
```diff
@@ -15,36 +17,58 @@ import (
...
+	return cmd
+}
 
-	if printVersion {
-		fmt.Fprintln(os.Stdout, kpcommon.Version())
-		os.Exit(0)
+func main() {
+	cmd := newRootCommand()
+	cmd.SetArgs(normalizeLegacyArgs(os.Args[1:]))
+	kpcommon.ExecuteWithFang(cmd)
```
ioerror, line 43: Note: this changes courier -v / http-proxy-client -v from a bare version line to Fang's formatted output, which is the consistency we want but will break anything parsing those two.

gabrix73:
Thanks for the detailed review. I addressed the requested changes in `6e7ce082`:

- moved the legacy long-flag handling into the shared `common.NormalizeLegacyLongFlags` helper;
- made normalization position-aware, including flag values and the `--` terminator;
- added regression tests confirming that legacy flags reach each command's `Run` callback.

The Fang-formatted `-v` output remains intentional for consistency with the other commands.

Focused tests, race tests, compile-only checks, `go vet`, and `git diff --check` pass locally. The new CI run is in progress.

Please take another look when convenient.

ioerror:
Hi @gabrix73 - happy to review and thanks again for hacking on Katzenpost. Your prompt reply is most welcome!

ioerror (approved):
This is even better than my suggestions; thank you!

PR #1137, opened 2026-09-15
docker: cross-version interop gate for the testnet

ioerror:
Adds cross-version verification to the docker testnet: build releases alongside HEAD, swap some nodes to them, and confirm consensus still forms and a client still round trips. Additive `make` options; existing targets behave the same. The only behavior change is in CI: the gate runs per PR, goes faster on client-only changes (checks the live namenlos net instead of a full boot), falls back to the docker testnet if that can't validate, and is skipped for changes unrelated to the testnet.

Demos:

```
make interop-gate                                   # HEAD vs latest release (3/3 authority split)
make interop-gate release_refs="v0.0.103 v0.0.104"  # HEAD vs two releases (2/2/2 split)
make client-check                                   # working-tree client vs live namenlos
make net_name=a interop-gate & \
make net_name=b release_refs="v0.0.103 v0.0.104" interop-gate & wait   # in parallel, different versions
```

Default topology now mirrors namenlos (6 authorities, 8 mixes, 5 gateways, 4 service nodes, 4 storage nodes, with courier and echo on every service node), matching or exceeding it on every axis; `make start`/`wait`/`run-ping`/`test` are otherwise unchanged.

Minimal alternative to #1109.


ioerror:
This branch's main feature is that it is able to tell us if it is safe to tag HEAD when HEAD is run against the current (and/or previous) version deployed in the network.

tirewapped:
note since i see @ioerror is soliciting reviews for this on other PRs i should note there were several unresolved bits of feedback i gave him out of band: i don't understand the rationale for allowing a 4 of 6 consensus to pass; this seems to undermine the benefit of cross version checking. i don't think we should have a cronjob for running the docker tests (as discussed at length in #1171) and i do think we should have one for namenlos (and i don't think PRs should block on namenlos tests). there were probably more but i'm going offline for the evening now.

ioerror:
> note since i see @ioerror is soliciting reviews for this on other PRs i should note there were several unresolved bits of feedback i gave him out of band:

Thanks for leaving the feedback here. If possible, please leave a review using the review tool and then we can discuss exactly the code or other change that is at issue. It is much less confusing and allows us to systematically track and resolve the issues raised.

> i don't understand the rationale for allowing a 4 of 6 consensus to pass; this seems to undermine the benefit of cross version checking. 

The number of votes of a dirauth scale (somewhat) for the number of versions tested as shown here:
```
make interop-gate                                   # HEAD vs latest release (3/3 authority split)
make interop-gate release_refs="v0.0.103 v0.0.104"  # HEAD vs two releases (2/2/2 split)
make client-check                                   # working-tree client vs live namenlos
make net_name=a interop-gate & \
make net_name=b release_refs="v0.0.103 v0.0.104" interop-gate & wait   # in parallel, different versions
```

For the 4/6 example: Consider the case of two known good versions, and one questionable version where the dirauth and a mix server were changed in the questionable version. We would want the consensus to form with 4/6 dirauths, and then the possible problematic mix server would be in the consensus which formed due to the two good versions. 

Now consider a 3/3 example: one known good version, and one possibly problematic version, if the dirauth is problematic, it won't form a consensus at all.

What is the alternative suggestion? I am open to fixup commits that center around a clearly described test case.

What about that design rational isn't understandable? The consensus forming is important stuff but n/x seems to either need to match namenlos (4/6) or it should have a different operator/developer story. 

Regardless of the threshold, it is only a part of what we want to test regularly. The changes expected in the future include some additional dirauth changes but primarily it will cover other issues. What number or change would you propose and where specifically do you want to see it set to something else?

>i don't think we should have a cronjob for running the docker tests (as discussed at length in #1171)

It makes sense to me to regularly have HEAD test against the known good version(s) running on the network. This cron job would ensure that HEAD is always able to be tagged for release and that it will work with the currently deployed version. How do you propose that we do that? Currently, we just ask people to upgrade if the docker testnet works. If operators don't upgrade all at once, we accidentally test this kind of situation on the real namenlos network. Isn't it better to continuously test this kind of interoperability? It helps to test this regularly as changes happen and before operator's time is wasted.

> and i do think we should have one for namenlos (and i don't think PRs should block on namenlos tests).

That is orthogonal to this interop branch as of yet.  Note, #1109 was closed in part to make this very specific feature smaller. If you're requesting I add namenlos CI tasks as I did with katzenqt recently, I am willing to add a similar CI configuration. Please be specific about what you'd like to see. I will attempt to add similar commits that enlarge the scope of the changes in this pull request to address your feature request.

gabrix73:
**@ioerror** thanks, this thread clarifies the intent. Two concrete, non-blocking suggestions that align with @tirewapped's namenlos point:

1. Keep the interop gate off the docs path: a docs-only change currently triggers the full `interop` job. Restricting the gate to Go source changes keeps contributor feedback loops fast without weakening the check.
2. Keep the namenlos lane out of the blocking PR path. Exercising the public mixnet from GitHub runners couples PR outcomes to an external network neither the test nor the fixture controls. A scheduled cron for the namenlos lane (like the gateway status check) would still give the continuous interop signal, while the docker(compose) testnet stays the blocking gate.

The `startResending` hunk-order note in the stacked PRs remains a merge-order detail, not a blocker.

ioerror:
> **@ioerror** thanks, this thread clarifies the intent. Two concrete, non-blocking suggestions that align with @tirewapped's namenlos point:
> 
>     1. Keep the interop gate off the docs path: a docs-only change currently triggers the full `interop` job. Restricting the gate to Go source changes keeps contributor feedback loops fast without weakening the check.
> 
>     2. Keep the namenlos lane out of the blocking PR path. Exercising the public mixnet from GitHub runners couples PR outcomes to an external network neither the test nor the fixture controls. A scheduled cron for the namenlos lane (like the gateway status check) would still give the continuous interop signal, while the docker(compose) testnet stays the blocking gate.
> 
> 
> The `startResending` hunk-order note in the stacked PRs remains a merge-order detail, not a blocker.

Okay, I've tried to address this feedback by @gabrix73 and also earlier comments by @tirewapped:

- In 03045e5 fixes the fallback so that the interop runs only when server paths changed, or when a client change ran namenlos and it did not succeed. A README-only PR should now skips both and the gate should pass on the changes job alone.

- In 3925606 the client-check on namenlos moves to a new scheduled workflow, six-hourly offset by thirty minutes from the docker-mixnet cron, plus workflow_dispatch, with no continue-on-error since a red run is the signal. The PR gate drops the namenlos job entirely which is basically what @tirewapped implied but didn't precisely say; I expect this won't be enough but I am hoping it is what he wanted too. Interop needs only the changes job and runs on server changes. The gate requires the local -short lane and the interop lane to pass when they ran. The README paragraph describing the old routing was also updated, I hope it makes things clearer.

gabrix73:
Confirmed, the two follow-up commits address our feedback:

- the docker interop lane is now gated on server-path changes (plus the client-path case still covered by the scheduled namenlos check), so a docs-only PR no longer triggers the full interop run;
- the namenlos client check is moved off the PR gate into the scheduled `namenlos-client-check.yml` workflow, keeping the public-mixnet dependency out of CI merges.

Thanks for the fast turnaround.

ioerror:
@gabrix73 Glad to hear it, thank you. Unfortunately GitHub doesn't count your approval as a full approval (0.5 approvals out of 1 needed to merge, apparently), so this is still blocked on reviewer approval.

david415:
## Automated review

**Bottom line: I would not merge this as-is.** This changes the project's CI safety net, and two of the findings below are live regressions in test coverage guarantees for every future PR, not just this one. Both are small, mechanical fixes.

### Blocking

1. **`.github/workflows/docker-interop-gate.yml:30`** — The `changes` job's path filters omit several packages docker integration tests actually depend on, most notably `pigeonhole/` (the shared protocol layer used by client, courier, and replica per this project's own architecture docs), plus `common/`, `quic/`, `bacap/`, `loops/`. A PR touching only `pigeonhole/*.go` matches neither the `client` nor `server` filter, so both new jobs report `skipped`, which the `gate` job (`if: always()`) treats as passing. **A PR can merge fully green with zero docker-mixnet integration coverage**, exactly where the old mandatory full-mixnet PR gate used to sit.
2. **`.github/workflows/docker-interop-gate.yml:68`** (and `:109`) — Both new jobs run with `dockertest_race=` (explicitly emptied), disabling the Go race detector on every PR that touches client/server code. The old gate (`docker-mixnet.yml`'s `make dockertest-all`) hardcoded `-race`. Concurrency bugs in exactly the code paths this project's own concurrency-pattern docs are written for now merge undetected until the next 6-hourly scheduled run at the earliest.
3. **`.github/workflows/docker-mixnet.yml:2`** — The `pull_request` trigger is dropped entirely; combined with findings 1 and 2 above plus weaker failure diagnostics (below), the replacement is not equivalent to what it replaces, despite the project's own docs still describing `docker-mixnet.yml` as the full-suite PR gate.

### Correctness of the interop comparison itself

4. **`docker/Makefile:974`** — `interop-ref-binaries` hand-respells its `ldflags` instead of extending the shared `ldflags` variable every other build recipe uses (even the `-extldflags` syntax was retyped). If `${ldflags}` ever changes, reference binaries silently stop matching HEAD's build configuration, undermining the validity of the interop comparison with no error.
5. **`docker/Makefile:959`** — `docker_run_ref_sh` reimplements `docker_run_sh`'s container args (cache mounts, `GOCACHE`/`GOPATH`, `--pids-limit`, network mount) by hand instead of wrapping the shared variable — same silent-drift risk as above.
6. **`docker/Makefile:964`** — The per-ref build loop in `interop-ref-binaries` has no `set -e`; a failed build for one ref can be masked by a later ref's success, and `make interop-ref-binaries` reports success anyway.
7. **`docker/Makefile:966`** — The cache-hit check only tests `server.$(distro).$$ref` and infers `dirauth`/`replica` are also cached; the shared cache dir has no locking despite the PR's own README documenting running multiple `interop-gate` invocations concurrently with overlapping `release_refs`.

### Topology size jumped ~80%, supporting tooling didn't follow

8. **`tools/chaos/chaos.go:14`** — Still hardcodes the old topology (3 auths, 1 gateway, 3 mixes, 5 replicas). `Validate()` now rejects legitimate new node names, and the two shipped example chaos scenarios both target `replica5`, which no longer exists by default.
9. **`docker/Makefile:533`** — `watch-gateway`/`watch-all-separate` still hardcode `gateway1` only (not a glob, unlike the mix/auth/replica/servicenode watchers); `courier-restart`/`servicenode1-restart` only ever touch `servicenode1`, so 3 of the new default's 4 service nodes silently keep stale code after a "restart."
10. **`docker/Makefile:69`** — CLAUDE.md's documented defaults for `serviceNodes`/`storageNodes` (3/5) no longer match the new 4/4 defaults, in two separate places in that doc.

### Robustness

11. **`docker/client-configs/namenlos.toml:7`** — The `client-check` sed rewrites of the abstract socket name and `$XDG_RUNTIME_DIR` path have no match verification (`sed` exits 0 on zero substitutions); if either literal's spelling ever changes, the rewrite silently no-ops and the daemon either collides with a real katzenqt daemon's default socket or fails to start.
12. **`docker/Makefile:147`** — `compose_project_name`'s ordering hazard relative to `base_port`/`.base_port_seed` is silenced with `2>/dev/null` rather than fixed — more likely to matter now given the concurrent-`interop-gate`-invocations workflow this PR's own README documents.
13. **`docker/Makefile:463`** — Pre-existing bug sitting right next to this PR's own edit to the same line: the `cd $(net_name)` earlier in the `stop` recipe persists into the trailing courier-socket cleanup glob, which resolves to a doubly-nested nonexistent path and never matches anything.

### Test/CI quality

14. **`client/pigeonhole_docker_test.go:135`** — 13 duplicated `if testing.Short() { t.Skip(...) }` blocks with no shared helper; the fast lane is exclusion-based (a new slow test runs by default unless its author remembers the boilerplate).
15. **`.github/workflows/docker-interop-gate.yml:56`** — The `local` job (gating the highest-traffic client-path changes) has no cache step and a 25-minute budget for building the full 27-node topology, versus `interop`'s cached 120-minute budget for comparable work — it's simultaneously the most cache-starved and most time-constrained lane.
16. **`.github/workflows/namenlos-client-check.yml:1`** — No caching, 4x/day cron, full cold build every run, plus a fixed `sleep 15` instead of readiness polling before dialing the daemon.
17. **`.github/workflows/docker-interop-gate.yml:71`** — Failure diagnostics reduced to a 200-line kpclientd log snapshot only; the old workflow additionally dumped replica logs and uploaded a full testnet-state artifact.

ioerror:
Thanks @david415, I pushed changes that should address all of your concerns:
- 1, 3: the local job runs on any Go change and the server filter covers the shared packages; the full suite is on the schedule and release tags by design, and the README is updated to state which job runs what, etc.
- 2, 15, 16, 17: -race is back on both jobs, both are cached, and both dump replica logs and upload the testnet state on failure; the namenlos check is cached and polls for the daemon socket. Note that windows had those removed (see the recent PR on main that was merged) due to golang itself having a crash bug on windows when -race was used.
- 4 to 7: reference binaries use the shared ldflags and container args, the loop fails on the first bad ref, all three binaries must be cached, and builds run under flock.
- 8, 9, 12, 13: chaos hosts and the watch and restart targets follow the topology, the port seed is evaluated once, and stop cleans up from the right directory.
- 10: I could not find your stated defaults documented in the tree; point me at your private rules file if you have something more specific in mind.
- 11: the socket rewrites are verified before and after the sed.
- 14: one skip helper; the job stays exclusion-based on purpose.

Also to make it explicit the testnet after merging will match the shape of namenlos (6 authorities, 5 gateways, 7 mixes, 4 service nodes, 4 replicas). The namenlos cron job pings the echo service after the fetch, and a daily workflow checks HEAD against the last two releases with mixed authorities and with every node on HEAD under old authorities. It is also possible to run many mixnets in parallel now - this along with the version means that on a large machine, we could probably run every version combination to understand the full compatibility matrix. We don't need to do such a large job, of course, but it is nicer for local development - I run one testnet for katzenqt work, and one testnet for katzenpost; they don't stomp on each other and i am able to easily glance and see which testnet belongs to which development task.

The included cronjobs are useful to the project. Why? As one small example, we want to know answers to these questions without blocking developers or making the CI _less_ reliable:
- does kpclientd on HEAD (or any tagged release) currently work against namenlos?
- does a testnet function (form consensus, route traffic, run long running tests that span epochs) when HEAD is mixed into the existing tags?
- are there flaky tests that are unrelated to pull requests under discussion when we gather data over time?
- does all of that change or does it hold steady over time?

Please let me know if anything specific remains to be done, especially if until finished any remaining changes would otherwise remain a blocker for approval. I need to land this so that I am able to extensively test the dirauth changes before we tag a release, and before I land a new dirauth after the next tag. Thanks again for your review.

gabrix73 (approved):
Approving at head c20b6c73. The automated-review concerns raised by @david415 have been addressed in the follow-up commits:

- cross-version interop stays wired into the changes/gate chain and the local lane covers generic Go changes ("good to see the gate remains the single place to enforce cross-version tests" feedback is resolved);
- the race detector is enabled again in the gate and interop lanes;
- the docker-mixnet.yaml trigger dependency is preserved via the changes job.

All checks pass at this head, including `changes`, `gate`, `interop` and `local`. This matches the goal of landing an interop gate that exercises the changelog-controlled cross-version matrix before we tag a release and land the new dirauth. Thanks for writing the cross-version gate and for keeping the tests wired to the PR gate.

tirewapped:
> For the 4/6 example: Consider the case of two known good versions, and one questionable version where the dirauth and a mix server were changed in the questionable version. We would want the consensus to form with 4/6 dirauths, and then the possible problematic mix server would be in the consensus which formed due to the two good versions.

This seems to treat the 3 involved versions as equals, when in fact they are not: the latest is the most important one. If we want to say having old versions not in the consensus is acceptable, then we wouldn't care about cross-version interop. and obviously if the new version is not making it in the consensus, that should also be a blocker. And in any case, the most important thing is that the new version by itself can make consensus. I am still baffled as to why we would ever want a test which calls 4 of 6 a pass.

ioerror:
> > For the 4/6 example: Consider the case of two known good versions, and one questionable version where the dirauth and a mix server were changed in the questionable version. We would want the consensus to form with 4/6 dirauths, and then the possible problematic mix server would be in the consensus which formed due to the two good versions.
> 
> This seems to treat the 3 involved versions as equals, when in fact they are not: the latest is the most important one. If we want to say having old versions not in the consensus is acceptable, then we wouldn't care about cross-version interop. and obviously if the new version is not making it in the consensus, that should also be a blocker. And in any case, the most important thing is that the new version by itself can make consensus. I am still baffled as to why we would ever want a test which calls 4 of 6 a pass.

You are right that a threshold pass hides a broken authority, and that is fixed: every cross-version job now fails unless all 6 authorities sign and every mix, gateway, service node and replica is listed in the consensus, and the job prints one line per service with its version and whether it signed or was listed, so a broken authority and a broken node show up in the same run instead of one masking the other.

The dirauth is only part of it. Some namenlos operators (as you probably know...) run a single mix, a gateway, a service node or a storage replica, and they upgrade whenever they upgrade, so HEAD breaking any of those roles under old authorities, or old nodes under HEAD authorities, is a failure we need to see as much as an authority that cannot sign. That is why the daily workflow runs eight combinations, each one make target, each ending in 6 of 6 signatures, every node listed, every swapped service verified to run the requested revision, and pings passing:

- interop-authorities-mixed: authorities on HEAD, n-1 and n-2, nodes on HEAD (authorities upgraded one at a time)
- interop-authorities-old: authorities on n-1 and n-2 only, nodes on HEAD (nodes upgraded first)
- interop-authorities-old-n1: authorities on n-1 only, nodes on HEAD
- interop-nodes-mixed: authorities on HEAD, nodes on HEAD, n-1 and n-2 (nodes upgraded one at a time)
- interop-nodes-old: authorities on HEAD, nodes on n-1 and n-2 only (authorities upgraded first)
- interop-nodes-old-n1: authorities on HEAD, nodes on n-1 only
- interop-everything-mixed: everything interleaved across HEAD, n-1 and n-2 (mid-rollout, no ordering)
- interop-everything-mixed-n1: the same with one release back

HEAD against HEAD only is every pull request's docker run and the scheduled full suite, so it is not repeated here. Any of these runs locally with the same make target (at the cost of adding a handfull of Makefile make targets) in an effort to not lock us into github's runner config format.

With the assumption that the CI passes... @tirewapped is there anything else that needs to be done on this branch to land it?

ioerror:
@gabrix73 @david415 @tirewapped: In order to optimize the hot path, I've pushed another couple of changes.

When someone opens a pull request or pushes to to that pull request: the unit tests run as always, and one docker job builds a testnet that is roughly namenlos shaped but with a @tirewapped exception: 6/6 dirauths must sign for it to be successful. On namenlos, a consensus is formed with a simple majority. The namenlos-shaped testnet is entirely composed of nodes from HEAD and it runs the short client suite with the race detector enabled. This is about two minutes of tests after the build; the docker build itself could probably be better optimized but that is generally true and this branch doesn't change that outstanding concern in a meaningful way... So that is the whole fast path. Nothing on a pull request compares HEAD with old releases.

We also want to be able to manually dispatch jobs, and we want the full matrix of versions tested as well, so both are now supported. The deep testing happens on the (cron) clock: the full client suite every six hours and on release tags, and the eight-way upgrade matrix (HEAD against n-1 and n-2 in every operator ordering) once a day at 04:00 UTC. Anyone can run either by hand from the Actions tab (see the GitHub web ui above, click click), or locally with make to ensure that we aren't tied into the GitHub CI world. The make target follows the pattern of `interop-<job>` and also has a `make interop-matrix` target as a shorthand.

If a developer wants the cross-version check on their pull request, they _may_ add the interop label; that _intentionally forced_ interop job runs one extra job. This means that everything will be tested interleaved with the last two releases and uses 6 of 6 signatures (as requested by @tirewapped).  Here's an example of how to trigger the larger run when you want to: `git commit -m "server/pki: do something interesting [interop]" && git push origin feature_branch_foo_interesting`.  This stays out of everyone else's way unless it is explicitly requested but it does regularly run to show us if HEAD is ready to rumble on the real network...

Is there anything else that needs to change for this to be merged?

ioerror:
#1176 is now ready to merge into this pull request.

ioerror:
The CI run after merging the pull request from @gabrix73 into this branch demonstrates the new CI flow.

PR #1138, opened 2026-09-15
authority/voting: review the voting directory authority

ioerror:
Reviews the voting directory authority (the "og" dirauth) and fixes what
turned up.

Reject bad input instead of panicking (Addresses, short descriptor identity
keys). Guard the shared-random and certificate paths: set-commit, cert epoch
binding, vote-as-cert, reveal bounds, authenticator AD. Give the PKI wire
message a real size ceiling that covers the inter-authority certificate,
derived from the configured schemes and topology, and wire it into clients.

Connections: recover per-conn panics, cap concurrent and per-peer conns, log
and drop unknown peers, bound the first command and keepalive writes, drain
accepted conns on shutdown. Persistent inter-authority connections stay
opt-in, default off.

Adds a scheme matrix and per-fix regression tests. Verified on a full mixnet
and a mixed patched/unpatched authority quorum.


david415:
● Code review(review · 15 findings)
core/wire/session.go
  ●  609 [correctness]    SendCommand's mutex only serializes each individual cipher operation, not the whole logical send, so concurrent SendCommand calls on one Session can still corrupt the wire stream.
authority/voting/server/wire_handler.go
  ●  394 [correctness]    serveAuthorityConn's KeepaliveTimeoutSec idle deadline is clobbered back down to ResponseTimeoutSec on every RecvCommand call, silently defeating the persistent-connection feature under default config.
authority/voting/server/state.go
  ● 1122 [correctness]    A panic during a reused persistent-connection round trip skips the pc.closeLocked() eviction every normal error path performs, leaving a possibly-corrupted cached session for reuse.
  ● 2132 [security]       onVoteUpload, onRevealUpload, and onSigUpload were not given the peer-identity-binding check this PR added to onCertUpload, despite the identical threat applying and onVoteUpload's own comment acknowledging it.
  ● 1119 [correctness]    The per-peer mutex serializing persistent-connection round trips is shared across all four voting phases (Vote/Cert/Reveal/Sig), so a slow send in one phase can silently block a later phase's send to the same peer past its own deadline.
  ● 2811 [correctness]    backgroundFetchConsensus dials and handshakes to peer authorities via an independent client.Config, completely bypassing state.peerConns, so a consensus gap-fill fetch can open a second live connection to a peer that already has a cached persistent one.
  ●  329 [efficiency]     The FSM logs consensus size by marshaling the document purely to discard the bytes, instead of populating the new serializedDocs cache with them, so the same document gets marshaled again later on the first GetConsensus of the epoch.
authority/voting/server/peer_conns.go
  ●  197 [correctness]    sendPeerKeepalives calls session.SendCommand directly with no panic recovery, unlike doSendCommand, so a panic here crashes the whole dirauth process.
  ●   85 [correctness]    dialAndHandshakePeer's address loop discards the informative dial-failure error when the LAST address in the peer's list fails to parse rather than fails to dial.
core/pki/descriptor_impl.go
  ●  150 [correctness]    The root-cause panic (hpqc's hybrid sign scheme slicing an input with no length check) is patched only at two dirauth intake points; ReplicaDescriptor.DisplayWithSchemes still panics unconditionally on the same class of input and is reachable, unguarded, from courier debug logging.
authority/voting/server/server.go
  ●  374 [correctness]    Server.halt() waits for the WaitGroup to drain before calling state.Halt() (which closes all cached persistent peer connections), so an outbound send stuck on a peerConn can prolong shutdown well past what the PR intends.
  ●  260 [simplification] listenWorker's connSem-acquire select includes a <-s.haltedCh escape branch that is not actually reachable during the graceful-shutdown window it was added for.
authority/voting/server/config/config_impl.go
  ●  193 [correctness]    Server.validate() no longer requires PKISignatureScheme to be set; it now silently substitutes a default instead of failing config load.
authority/voting/client/client.go
  ●  135 [correctness]    The client's derived wire-message ceiling assumes a hardcoded topology size cap (128 nodes, 32 replicas) rather than the real network, so a network that grows past that assumption gets a ceiling smaller than the dirauth's own topology-exact estimate.
authority/voting/server/shutdown_drain_test.go
  ●  100 [test-coverage]  TestShutdownClosesIdleAuthorityConn leaks the connection-handler goroutine and both net.Pipe endpoints if any assertion between spawning the handler and the end of the test fails.

Review of PR #1138 is done — 15 findings, most-severe first. Standouts:

- core/wire/session.go:609 — the new mutex on SendCommand only guards individual cipher ops, not the full send; concurrent callers on one Session can still desync header/payload encryption and interleave writes on the wire. The new race test only proves -race is quiet, not that it's protocol-safe.
- authority/voting/server/wire_handler.go:394 — the persistent-connection idle deadline (120s) gets silently clobbered back to the response timeout (30s) on every RecvCommand, defeating the PR's whole point under default config.
- state.go:2132 — onCertUpload got the new peer-identity-binding check, but onVoteUpload/onRevealUpload/onSigUpload didn't, despite onVoteUpload's own comment claiming that protection exists. A byzantine authority can replay another authority's vote/reveal/sig over its own connection.
- peer_conns.go:197 — sendPeerKeepalives calls SendCommand with no panic recovery, unlike every other send path this PR hardened — a panic here crashes the whole dirauth.
- core/pki/descriptor_impl.go:150 — the underlying hpqc panic-on-short-key bug is patched at two dirauth intake points but not in DisplayWithSchemes, which is still reachable unguarded from courier debug logging.

Plus several more: a shared per-peer mutex causing cross-phase head-of-line blocking, shutdown ordering that can stall behind a wedged peer send, a config validation that now silently defaults PKISignatureScheme instead of failing, a hardcoded topology-size cap in the client's message-size ceiling, and a couple of minor test/logging/error-message issues.

ioerror:
I have addressed the review in fixup commits - I am pushing back on only one of them: having a default signature scheme for the dirauth has already shaken out some bugs including on windows. We want to consistently have sane defaults such that no config file is required except for overriding the sane defaults. This is the same reason we have discussed unnecessary crashes when we can't push out a config file at the same time as an update. This is a failure mode that we can avoid as we have previously agreed to do so. Lets avoid that one change and otherwise the rest of your change requests are addressed.

david415 (approved):

ioerror:
I suspect that merging this will create overlap to reconcile against my other open pulls on these branches: dirauth-shutdown-drain, dirauth-fsm-sleep-clamp, wire-pki-message-ceiling, and dirauth-max-concurrent-conns.

PR #1139, opened 2026-09-16
courier: bound reply writes, recover the tombstone worker, guard WriteCap

ioerror:
Hardening for the courier's client-facing copy path, from a panic/DoS audit.

- Bound the per-request reply-write goroutines in OnCommand with a semaphore
  (256): a stalled socket can no longer grow goroutines unboundedly.
  Saturation drops the reply (the client's ARQ retransmits) instead of
  blocking OnCommand, avoiding head-of-line stalls.
- Recover panics in the detached writeTombstonesToTempChannel goroutine, which
  ran outside the copy worker's recover; a panic there no longer crashes the
  courier.
- Add a regression test that a crafted off-curve WriteCap is rejected with a
  failed reply, not a crash. The root fix is in hpqc (ed25519.FromBytes,
  katzenpost/hpqc#118); the test skips until that lands, so this merges now
  against the current hpqc and the version bump can be a commit on top.


david415 (approved):

ioerror:
This shouldn't have been merged before the CI finished but I mistakenly pushed the button.

PR #1146, opened 2026-09-16
connlimit: split inbound connections into client and peer pools

ioerror:
Cap inbound connections in two independent pools keyed by source IP: a client pool for non-peer IPs and a peer pool for mixnet servers/authorities from the consensus, each with a per-IP cap, so a client flood can't starve server-to-server links and a bad peer can't starve clients. Defaults are 1024/1024 (client/peer) and 64 per IP, centralized in core/connlimit and overridable per role; classification is by source IP against an epoch-updated peer set.


david415 (approved):
I added a minor fixup commit. I hope you don't mind.

ioerror:
> I added a minor fixup commit. I hope you don't mind.

I don't mind at all, thank you!

PR #1153, opened 2026-09-16
Fuzzing kit

ioerror:
A native go test -fuzz tiny tier and an autoscaling libFuzzer big tier built from the same testing.F targets via go-118-fuzz-build (fork mode, one worker per core, fills a 2-core or a 200-core box). Systematic parse-surface targets across the major components: wire commands per role, sphinx packet and routing, bacap caps, the PKI client and PKI server/consumer, and PKI documents and certificates.

Run one target overnight:

    make -C fuzz overnight FUZZ=./core/wire/commands/ FUZZFUNC=FuzzMixnetCommandsFromBytes

All fuzz code is behind //go:build fuzz, so normal builds and CI are untouched. Bugs reproduced from campaigns go to the fuzzing-findings branch (#1152).


ioerror:
This branch contains a basic fuzzing setup that helped me to find several issues (see #1152 ) on X86_64 and currently there is an outstanding run for s390x to ensure that we have both little endian and big endian runs for all the fuzzing targets.

david415 (approved):

PR #1154, opened 2026-09-17
Bump hpqc v0.0.88

david415:
breaking changes to hpqc's BACAP api now require us to change our BACAP call sites here.


ioerror (approved):
SignBox error paths are currently untested and I'll push a fixup and then merge.

PR #1161, opened 2026-09-21
pigeonhole: regenerate trunnel output for bulk byte-array copy

david415:
## Summary

Regenerates `pigeonhole/trunnel_messages.go` with katzenpost/trunnel master `c7f56069d31600356ef64933ec65ee56bf95584b`, which includes katzenpost/trunnel#3 (bulk copy of unconstrained `u8` and `char` arrays).

- Parse of fixed and length-prefixed `u8` arrays is a bounds check plus `copy` instead of a per-byte loop.
- `MarshalBinary` appends byte slices directly.
- The empty per-byte validation loops are gone.
- Parsed values still own their storage; bytes are copied, not aliased to the input buffer.

Only `pigeonhole/trunnel_messages.go` changes (79 insertions, 217 deletions). `gen-fuzz.go` and `gen-marshallers_test.go` regenerate byte-identical.

## Performance

`CourierEnvelope`, generated from `pigeonhole_messages.trunnel`, measured locally (single 300 ms run, illustrative payload sizes):

| Operation | Before | After |
| --- | ---: | ---: |
| Marshal, 1 KiB | 1412 ns, 9 allocs | 306 ns, 6 allocs |
| Marshal, 64 KiB | 65347 ns, 22 allocs | 4610 ns, 6 allocs |
| Parse, 1 KiB | 1107 ns | 182 ns |
| Parse, 64 KiB | 46049 ns | 4306 ns |

`ReplicaWrite` and `Box` show similar gains. Marshal still starts from an unsized buffer, so it still allocates more than once; that is a generator follow-up and is not part of this change.

## Verification

- Control: regenerating with the previous generator (`d46209b0be0e44aafba1087bb4a71b18f5d38b69`) reproduces the tracked files byte for byte, so this diff comes from the generator change alone.
- Differential test outside the repo, old versus new generated code over about 630,000 mutated inputs (truncations, length-field rewrites at the exact-fit boundary) for all 14 message types: identical errors, leftover bytes, parsed values (including nil versus empty slices) and re-marshalled bytes. The test was confirmed able to fail by injecting bugs.
- `gofmt`, `go vet ./pigeonhole/...`.
- `go test -race ./pigeonhole/... ./courier/...`.
- Replica integration tests with the CLAUDE.md CGO invocation: `TestCourierReplicaIntegration`, `TestCourierReplicaSequenceIntegration`, `TestProxyIntegration`, `TestReplicaReplication`.
- The docker mixnet tests were not run locally; CI covers them.

## Note

`pigeonhole/build.go` renames the generated test file to `trunnel_messages_test.go`, but the repo tracks it as `gen-marshallers_test.go`, so a plain `go generate` leaves the tracked file deleted and adds a duplicate. I renamed it back by hand for this regeneration and did not change `build.go` here.


tirewapped (approved):
out of an overabundance of caution i tried regenerating this myself (using our latest trunnel with https://github.com/katzenpost/trunnel/pull/3 landed, which i also lightly reviewed) and can confirm that i reproduced the code generation in this commit. those performance improvement numbers look great, thank you @gabrix73 !

PR #1170, opened 2026-09-22
ci: stop Go 1.26 race/coverage crashes on the windows unit tests

ioerror:
Go 1.26's race detector and coverage writer intermittently corrupt memory on Windows (found pointer to free object / unexpected fault address) after the tests already pass; this drops -race and -coverprofile on the windows unit tests only, so every PR stops flaking there while Linux keeps running both.


tirewapped:
this PR description says it just fixes the windows CI but it this also contains 77724f1e42495d022261b2e25c98acda7028d2b8 which is another copy of 4bbcf01d69f9b56318046f5fe0c02fdc735499de from #1171

ioerror:
> this PR description says it just fixes the windows CI but it this also contains [77724f1](https://github.com/katzenpost/katzenpost/commit/77724f1e42495d022261b2e25c98acda7028d2b8) which is another copy of [4bbcf01](https://github.com/katzenpost/katzenpost/commit/4bbcf01d69f9b56318046f5fe0c02fdc735499de) from #1171

Yes, that was a copy picked onto this branch so its own docker-mixnet run stopped flaking. I didn't put it into the PR but I would have been happy to amend the PR text as well.

In an effort to get these windows CI failures out of the way for the GNU/Linux development tasks, I have dropped the commit, and rebased to the two windows commits only which have been causing recent failures.

tirewapped:
thanks for removing the commit here from 1171.

i don't care so much about windows (and aside from kpclientd i don't expect anything in this repo is likely to be deployed there) so i'm tempted to just merge this but i am curious: are you suggesting that these race failures are specific to 1.26?

can/should we try upgrading to go 1.27.1 instead of doing this, or revert this PR when we do upgrade later?

ioerror:
> thanks for removing the commit here from 1171.

No problem. We can discuss the cron-like feature it in the other pull request.

> 
> i don't care so much about windows (and aside from kpclientd i don't expect anything in this repo is likely to be deployed there) so i'm tempted to just merge this but i am curious: are you suggesting that these race failures are specific to 1.26?
> 

They appear to be specific to windows. It may be specific to golang 1.26 on the github runner machines that ran them. I recall that the failures happened after the tests passed which seemed surprising.

> can/should we try upgrading to go 1.27.1 instead of doing this,

For windows only or for the entire repo?

> or revert this PR when we do upgrade later?

I don't feel strongly about it. I only want the windows crashes to not block unrelated work. This pull request set of changes stops the failure on 1.26.x. Either way the windows tests were (and are) passing.

tirewapped (approved):

PR #1171, opened 2026-09-22
ci: run docker-mixnet on schedule and tags, not per-PR

ioerror:
The full docker suite flaked on every PR; run it 4x daily and on release tags instead. Matches the docker-mixnet trigger in #1137, so they merge cleanly in either order.


tirewapped:
what is the process for surfacing failures from the scheduled runs, and/or avoiding finding out that a tagged commit fails only *after* the commit is tagged?

ioerror:
The purpose of this is to ensure there is any long running CI job at all - currently, it will happen on a regular schedule and at tag time - it is the final test of a cut release but hopefully it is not the only time this happens. If it fails for a serious reason, then we should not use that version. The long running CI jobs are blocking work and they are too flaky to run on ever push or pull.

ioerror:
> what is the process for surfacing failures from the scheduled runs, and/or avoiding finding out that a tagged commit fails only _after_ the commit is tagged?

We receive an email if and when the CI jobs fail. We can also see it in the web interface. When cutting a release, we do not currently have any automated QA that ensures that everything is tested as well as it could be tested. This is a useful datapoint. I expect that the only time a tagged commit will fail is if that is the first and perhaps the only time the branch has ever been tested. The exception is for really flaky tests but hopefully we will continue to solve those as they arise.

We could consider something additional such as running some kind of (katzenqt) bot to surface it in a development channel.

How do you normally see CI runs fail and why do we need something more?

tirewapped:
so we find out main is broken only when a cronjob runs or a tag gets tagged? and then we don't know which of the last N PRs might have broken it?

doing a git bisect over potentially many PRs to figure out which one broke CI sounds like no fun. this is why i want it to run on every PR.

it would be nice if it didn't necessarily need to run on every push, but at least on the final commit which is going to land it should run before the PR is merged.

ioerror:
> the notion of finding out main is broken because a cronjob ran or a tag got tagged, and not knowing which of the last N PRs might have broken it is horrifying. 

The only thing worse is not knowing at all. Regular runs will ensure that we also fix flaky tests and improve performance of the network itself. This kind of failure is not always evidence that changes between are bad, sometimes it is evidence that we need additional changes that we hadn't considered. The flaky tests are a real issue and we'll solve them faster when we can show that they happen outside of any single pull request.

> doing a git bisect over potentially many PRs to figure out which one broke CI sounds like no fun. 

In the case of a flaky network, we wouldn't need to git bisect. In the case of a release missing a feature that needed to land for the release to be functional, we also wouldn't need to git bisect. We have run into both in practice. In the case that a release is broken, a check after a tag is cut may indeed lead us to need to git bisect to resolve it. Better is to know.

>i it to run on every PR.

There is a tension here. I generally agree with you. That is probably the job of the author and likely also the reviewer. Do you know how to make that happen in a way where we could find common ground?

> it would be nice if it didn't necessarily need to run on every push, but at least on the final commit which is going to land it should run first.

That is exactly the tension right now. It is currently very slow and it blocks useful work because we have limited CI capacity.

tirewapped:
> The long running CI jobs are blocking work and they are too flaky to run on ever push or pull.

This sounds like an argument for running the tests *less*, but

> The only thing worse is not knowing at all. Regular runs will ensure that we also fix flaky tests and improve performance of the network itself.

this is an argument that we... should be running them *more*?

Adding a cronjob means regularly re-running them on the same code, and dropping the push trigger means running them on far less different commits. Retesting the same thing automatically only really makes sense to me for finding *rarely*-flaking tests; for frequently flaky tests we'll see the failures anyway and know we need to fix them.

imo we should strive to avoid having main fail CI. if we require CI to pass before merging, then in theory the only way CI should fail on main is a flaky test (so, it should pass with enough re-runs). it becomes quickly clear which tests are flaky and they can/should be fixed.

> Do you know how to make that happen in a way where we could find common ground?

Regarding not running on every push, I think on github we can make a workflow be required but not automatically run? Then someone would need to click a button to manually trigger it to make their PR mergeable. I'm not sure. (also imo it is fine to bypass this merge requirement sometimes if someone is making a docs-only change or another change which they are very sure won't break CI.... but then we should have it run on every commit that is the tip of main also so it is unambiguously clear to the next person that they are starting from green ci..)

ioerror:
> > The long running CI jobs are blocking work and they are too flaky to run on ever push or pull.
> 
> This sounds like an argument for running the tests _less_, but

It is an observation that due to a lack of resources, we should at least try to run them at different times and especially when it really matters. A tagged release almost certainly needs a bare minimum amount of automated testing before tooling or users attempt to use it. Regular runs of long running tests (sometimes over an hour or more) for tags and on a regular time basis are a simple QA tests that as of now is more than we are currently doing. Neither is harmful.

> 
> > The only thing worse is not knowing at all. Regular runs will ensure that we also fix flaky tests and improve performance of the network itself.
> 
> this is an argument that we... should be running them _more_?

In a non-blocking way, yeah?

Running various tests on a regular basis is helpful. The specific (cron like) timing is largely irrelevant to me and I am open to a change that picks different values if that is the sticking point.

What is unfortunate is that currently the CI blocks developers from iterating quickly. The CI jobs can last over an hour for every push and every PR even while a branch is under active development. This may lead to tons of irrelevant and queued jobs that block other CI runs for other pull requests or it may lead to canceling outstanding jobs, or some combination of the two.

This PR change also enables manually dispatching which is sorely needed. An author and a reviewer under these changes are now also able to trigger a workflow manually.

> 
> Adding a cronjob means regularly re-running them on the same code,

Yes, the same code and this is good because not everything in the run is certain to be the same.

> and dropping the push trigger means running them on far less different commits

So, if I add the push trigger back, are you okay with the `workflow_dispatch:` part?

> Retesting the same thing automatically only really makes sense to me for finding _rarely_-flaking tests; for frequently flaky tests we'll see the failures anyway and know we need to fix them.

The regularity could be once a week, it could be four times a day. Regardless, without this, we are primarily running on a pull request, a push to a pull request, and on a merge to main, and a few other situations but in most if not all cases it is extremely slow.

Manually triggering jobs requires `workflow_dispatch:` and regularly run jobs requires the cron like config. In a failed state, an authorized person is also able to  use the web interface to click the `re-run` button which in practice allows for masking intermittent failures.

A consequent of focusing on pull requests and merges (to main or elsewhere) being the primary CI testing trigger method is that a flaky test that lands on main _may not fail_. Unfortunately, the next pull request or merge to main could cause it to fail.

Regularly running it is not a problem, and it costs us nothing. It will almost certainly surface long neglected flaky failures outside of the context of a specific pull request. Is it a problem or does it cost us something that I am missing?

> imo we should strive to avoid having main fail CI.

Absolutely! Still, we will see that there is a failure on main without having to rely on each pull request. As an example: In the last weekend, we had many pull requests (on katzenqt and on katzenpost) merged. Some of them were merged before they were seen by more than two people (the author, and the reviewer). It would be helpful to have and to see CI runs for git tip of main outside of the context of any additional change or any pull request that may come later on top.

Relatedly, it would be nice if the docker testnet ran for many epochs rather than just one or two, for example. There exists a set of things that we may want to do regularly but where we don't want to have to manually trigger it. It is also nice to see consistent, repeated runs. At the moment, I think we would would see inconsistent, embarrassing failures. I would like to surface that kind of issue rather than paper over it or have the initial blame fall on a random person opening a PR.

> if we require CI to pass before merging, 

Yes, we agree. I would frame it as: No bugs is the goal and perfectionism is the ideal, the CI will give us a helpful reality check on how we're doing in reality.

>then in theory the only way CI should fail on main is a flaky test (so, it should pass with enough re-runs). it becomes quickly clear which tests are flaky and they can/should be fixed.
> 

Other options include a dependency changing out from underneath us (such as a docker runner image using a label that isn't a version number) and crash bugs that don't happen every time.

While I tend to agree that it should become clear which tests are flaky, this pull is attempting to run regularly to isolate those flaky tests and in a way that makes them independent of a pull request or a merge. It also enables a manual run that means anyone who is authorized will be able to trigger it at any time without having to push something.

> > Do you know how to make that happen in a way where we could find common ground?
> 
> Regarding not running on every push, I think on github we can make a workflow be required but not automatically run?

One part of that is to ask developers to put something in their commit message such as [skip ci] (see https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs ). Setting that up would mean that a push wouldn't trigger the CI if the developer adds [skip ci] to their commit.

For the other part, https://docs.github.com/en/pull-requests/reference/status-checks might be part of the answer.

I am not sure if any of that is what you would like to see, what do you think?

> Then someone would need to click a button to manually trigger it to make their PR mergeable. 

By asking developers to signal skipping in commit messages and ensuring the last commit doesn't skip, we wouldn't need a manual button push.

> I'm not sure. (also imo it is fine to bypass this merge requirement sometimes if someone is making a docs-only change or another change which they are very sure won't break CI....

That is easy to set for docs changes in a path that doesn't contain code and then there would be no required commit message either.

>  but then we should have it run on every commit that is the tip of main also so it is unambiguously clear to the next person that they are starting from green ci..)

That doesn't resolve the fact that right now, git tip of main can be green but on the next run, even in a branch from git tip of main, it will be red due to a flaky test or another latent reason.

It appears that we largely agree but what is missing for me is that your pushback is not clearly actionable. We are currently dealing with tons of flaky tests. They aren't getting fixed without a lot of friction, and getting more data by regularly running the CI isn't harmful.

If the only sticking point is that we should _also_ run on pull requests, I could live with that outcome. One outcome of keeping that is that we'll still see people manually cancelling jobs that are wasting time until the one that actually matters is scheduled to run.

I am open to compromises, would you be willing to push something that represents a constructive suggestion that we could merge based on our discussion so far?

tirewapped:
I think we're conflating multiple things here. We used to have branch protection rules which required CI to pass (or the bypass rules checkbox be checked) and we don't have those anymore since I'm not sure when. The current status quo is that we run ci automatically, and ostensibly don't merge things without a green checkmark... but sometimes we do.

I am fine with making the docker integration tests be `workflow_dispatch` (available at a button push) and not run on `push` but I am not fine with merging things where integration tests don't pass. I think we should re-enable the branch protection rules to require it (though it can still always be bypassed).

As I've said elsewhere I don't want to require github for contributions at all - I think we should be able to push merges to main as long as two regular contributors have signed off on the merge. I think our policy should be that those too should not break the docker integration tests, though it isn't automatically enforced. But for github we can and should automatically enforce it.

Running docker integration tests in cron is orthogonal and I don't see the point - as I said above it primarily would help surface rarely-flaky tests.

It would make sense to me to run the *namenlos* integration tests in ci cron, though, as that would be testing something which is more likely to have varying results from one run to the next.

tirewapped:
> but I am not fine with merging things where integration tests don't pass

note that this is generally-speaking, and of course there can be exceptions. eg if there is a known flaky test which was already failing it doesn't necessarily need to be fixed before anything else can land.

ioerror:
> I think we're conflating multiple things here.

We didn't resolve the issues related to this pull request's single commit. For example running this CI config on pushed (e.g. `v0.0.105`) tags aspect wasn't addressed by your replies, was it?

> We used to have branch protection rules which required CI to pass (or the bypass rules checkbox be checked) and we don't have those anymore since I'm not sure when.

What seems to be missing? Above I see that a review is required to merge. It also says that all checks have passed and so I think that means any requirements on the CI are met. Nothing in this pull request changes either of those things.

> The current status quo is that we run ci automatically, and ostensibly don't merge things without a green checkmark... but sometimes we do.
> 

Yes, unfortunately that has happened. I did that yesterday when a positive review was submitted but before the CI was passed. It was previously passing the CI but still, it wasn't the ideal moment to touch the button. I left a note in the pull request where it happened. Sorry again.

> I am fine with making the docker integration tests be `workflow_dispatch` (available at a button push) 

Okay, sounds good. Thank you.

> and not run on `push` 

So, no changes to my commit are required? These changes are only for the docker-mixnet CI config.

> but I am not fine with merging things where integration tests don't pass. I think we should re-enable the branch protection rules to require it (though it can still always be bypassed).

I generally agree but what does this have to do with my single commit in this pull request for the docker-mixnet CI config?

> 
> As I've said elsewhere I don't want to require github for contributions at all

This is orthogonal but nevertheless,  we agree. I run the CI locally in addition to using github's CI.

At the most recent KP meeting, we agreed that the only thing landing on git main is merge commits and tags.

However, I note that there was also discussion about how we do everything through pull requests which doesn't square with not requiring github as only on github do we have pull requests.I didn't advance the idea that we must do everything through pull requests but I don't object in principle. Probably we should move off of github but that is also orthogonal to this pull request.

> - I think we should be able to push merges to main as long as two regular contributors have signed off on the merge. 

We did not specify how they arrive on git main (e.g.: github pull request web ui, github cli tools, git push to main, etc), and we agree that pushing to main is something we should be able to do when the review requirements are otherwise met.

> I think our policy should be that those too should not break the docker integration tests, though it isn't automatically enforced.

In principle, I am in agreement with your policy idea. The forum seems a bit off for CI policy generally. I do not know what enforcement looks like here other than the git remote not accepting a branch on main that doesn't have a corresponding github CI run that is passing.

In the case that you push to git main and there is no pull request, no change made here impacts your proposed policy.

In the case that there is a pull request, no change made here impacts your proposed policy.

> But for github we can and should automatically enforce it. 

That policy has nothing to do with this commit. I don't feel comfortable setting the policy in this pull request as only the two of us in any case. I don't feel strongly about any of the policy suggestions, and I generally agree with your policy suggestions.

> Running docker integration tests in cron is orthogonal and I don't see the point - as I said above it primarily would help surface rarely-flaky tests.
> 

It helps surface constantly flaky tests and again, I don't see a downside. 

> It would make sense to me to run the _namenlos_ integration tests in ci cron,

That would be in a different ci config file that is not touched by this CI config, right?

The namelos tests are in the katzenqt repo. We do not have namenlos tests running on this pull request or in this repo as far as I recall.

>  though, as that would be testing something which is more likely to have varying results from one run to the next.

I would be fine adding a namelos CI test and I agree that it makes sense to run that in a cron-like CI config. Unfortunately, this too seems orthogonal to this set of changes in this pull request. There is no disagreement but shouldn't those namenlos changes go into a different pull request?

ioerror:
> > but I am not fine with merging things where integration tests don't pass
> 
> note that this is generally-speaking, and of course there can be exceptions. eg if there is a known flaky test which was already failing it doesn't necessarily need to be fixed before anything else can land.

Yes, we agree that there are exceptions. It is unfortunate that the exceptions are also the rule.

> if we generally require every merge to be passing before it lands, though, we should be able to relatively easily find the PR that introduced new flakiness (by re-dispatching it over the last few merges before the first observed failure and looking for another one) and the people (there are always at least two involved) who introduced that flakiness should be the ones to take responsibility for remedying it.

I generally agree. Lets put our agreement into action: we currently have long running tests that have lots of flaky failures in the docker-mixnet CI job. Should someone run git blame and file a ticket? Is that person going to be accountable for fixing it? What happens if it is inherent in the nature of the test or the resources running the test?

This pull request isn't really about accountability for flaky test fixing, it is about gathering and surfacing more data on a regular basis that shows the general state of things. It is also about trying to move the long running docker tests to not be in the hot path of a pull request. I don't feel so strongly about `-on: [pull_request]` - if you're asking me to keep that line, I am happy to push a commit that restores it.

tirewapped:
> For example running this CI config on pushed (e.g. `v0.0.105`) tags aspect wasn't addressed by your replies, was it?

It was. I suggested a superset of that: run it on every tip that is merged to main, and also on the resulting new tip of main. This way, main has always had it run, twice actually, and tags have it run *before* they're tagged.

> What seems to be missing? Above I see that a review is required to merge. It also says that all checks have passed and so I think that means any requirements on the CI are met. Nothing in this pull request changes either of those things.

Currently you need to check the "bypass rules" checkbox to merge without a github review (approval) but you don't to in order to click merge when the CI has failed. This is because (I just checked) under settings->branches->branch protection rules->main there are *no required checks*. We used to have many workflows listed there; possibly they got removed because you renamed the workflows? I'm not sure.

> I generally agree but what does this have to do with my single commit in this pull request for the docker-mixnet CI config?

By making them not run on push, this PR removes the red failure X which might dissuade someone from merging the thing that breaks CI. I am sorry for not being clearer but I think this is consistent with what i've been saying since my second comment: I am OK with removing the push trigger *but only if we make it a required check again* (so that people need to either trigger the workflow manually and see it pass, or click the red "bypass rules" checkbox).

> At the most recent KP meeting, we agreed that the only thing landing on git main is merge commits and tags.

Tags landing on main doesn't make sense, tags generally are put on a commit which is on main already (and typically the tip of main).

> It helps surface constantly flaky tests and again, I don't see a downside.

The downside is that it consumes our now-limited CI budget re-testing the same thing.

But whatever, if you insist on cron I'm OK with merging this for now as long as we can make the manually-dispatched workflow a required check. However, the autocomplete menu doesn't show any workflows matching "docker" though it shows our other ones as options :thinking: 

I also just noticed our current branch protection rule is now a deprecated thing and github says we should use their new rulesets instead: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets

i'm continuing to look into this.

ioerror:
> > For example running this CI config on pushed (e.g. `v0.0.105`) tags aspect wasn't addressed by your replies, was it?
> 
> It was. I suggested a superset of that: run it on every tip that is merged to main, and also on the resulting new tip of main. This way, main has always had it run, twice actually, and tags have it run _before_ they're tagged.
> 

That doesn't test tags. On what basis do you object to a CI run of the docker mixnet for a specific tag when it is pushed? I would like tags to have success or failure directly associated with the tag being pushed. This is useful.

> > What seems to be missing? Above I see that a review is required to merge. It also says that all checks have passed and so I think that means any requirements on the CI are met. Nothing in this pull request changes either of those things.
> 
> Currently you need to check the "bypass rules" checkbox to merge without a github review (approval) but you don't to in order to click merge when the CI has failed. This is because (I just checked) under settings->branches->branch protection rules->main there are _no required checks_. We used to have many workflows listed there; possibly they got removed because you renamed the workflows? I'm not sure.

I have not changed anything. Looking at all of the workflows, I am not the most recent person to touch any of them:
```
~/src/katzenpost/.github/workflows$ for f in `ls -1`; do git blame $f |tail -n 1; done
0b227b061d .github/workflows/benchmarks-daily.yml (David Stainton 2026-02-02 15:35:32 +0100 115) 
0b227b061d (David Stainton 2026-02-02 15:35:32 +0100 87) 
9f68df1bbe (David Stainton 2025-06-05 22:38:49 -0700 44)         run: cd ${{ matrix.test-suite.path }} && GORACE=history_size=7 go test -coverprofile=coverage.out -race -v -failfast -timeout ${{ matrix.test-suite.timeout }} ./...
d4fa774e30 .github/workflows/docker-mixnet.yml (Deckard         2026-08-08 01:00:35 +0000 100)             docker/mixnet-alpine/*/selfcheck.toml
6d731eb25f (David Stainton 2026-04-09 18:12:56 +0200 19)         run: go run golang.org/x/vuln/cmd/govulncheck@latest ./...
9f68df1bbe (David Stainton  2025-06-05 22:38:49 -0700 45)         run: cd ${{ matrix.test-suite.path }} && GORACE=history_size=7 go test -coverprofile=coverage.out -race -v -failfast -timeout ${{ matrix.test-suite.timeout }} ./...
4180324f56 (David Stainton 2025-06-05 19:26:51 -0700 36)         run: cd replica && GORACE=history_size=7 go test -coverprofile=coverage.out -race -v -failfast -timeout 30m ./...
9f68df1bbe (David Stainton 2025-06-05 22:38:49 -0700 78)         run: cd ${{ matrix.test-suite.path }} && GORACE=history_size=7 go test -coverprofile=coverage.out -race -v -failfast -timeout ${{ matrix.test-suite.timeout }} ./...
```

Between this merging and #1137 merging, a PR gets no docker-mixnet run at all. Note however that #1137 restores a per-PR docker run: its docker-interop-gate.yml triggers on pull_request and, when client code changes, runs the short lane (make dockertest_parallel=2 testargs=-short test, 20 min) plus the cross-version interop jobs; the full 2-hour suite stays on the schedule/tags. If you want a per-PR docker run continuously, merge #1137 first (it already carries the same docker-mixnet.yml trigger change) or merge #1171 and #1137 together.

> 
> > I generally agree but what does this have to do with my single commit in this pull request for the docker-mixnet CI config?
> 
> By making them not run on push this PR removes the red failure X which might dissuade someone from merging the thing that breaks CI.

As far as I understand the configuration the docker mixnet CI did not run on push to a branch.

> I am sorry for not being clearer but I think this is consistent with what i've been saying since my second comment: I am OK with removing the push trigger _but only if we make it a required check again_ (so that people need to either trigger the workflow manually and see it pass, or click the red "bypass rules" checkbox).

I think you are asking to _add a push trigger_ that is more than the pushing of a tag to trigger OR adding it as a required check. Both of which are fine with me but neither are something that I removed or changed. I removed only `on: [pull_request] ` in this pull request. You might be thinking about #1137, I suspect?

> 
> > At the most recent KP meeting, we agreed that the only thing landing on git main is merge commits and tags.
> 
> Tags landing on main doesn't make sense, tags generally are put on a commit which is on main already (and typically the tip of main).
> 

We discussed pushing tags over merge commits on main where the merge commit requires two people (author, reviewer) to approve while a tag over such a merge commit does not require two people to approve/push.

> > It helps surface constantly flaky tests and again, I don't see a downside.
> 
> The downside is that it consumes our now-limited CI budget re-testing the same thing.

That is a plausible motivation to fix the failures and to adjust the scheduled run time.

> 
> But whatever, if you insist on cron I'm OK with merging this for now as long as we can make the manually-dispatched workflow a required check. However, the autocomplete menu doesn't show any workflows matching "docker" though it shows our other ones as options 🤔

If you want to make it run once a week or once a month, I do not have a strong preference. I note that this enables the manual dispatch for the very first time.

> 
> I also just noticed our current branch protection rule is now a deprecated thing and github says we should use their new rulesets instead: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets
> 

That still doesn't explain how it was changed. I did not modify it, deprecated or not.

> i'm continuing to look into this.

Thank you.

tirewapped:
> That doesn't test tags. On what basis do you object to a CI run of the docker mixnet for a specific tag when it is pushed? I would like tags to have success or failure directly associated with the tag being pushed. This is useful.

A tag is just associating a name with a commit. If CI has already passed on that commit, then there is a CI pass associated with the tag from the moment the tag is created. Also, in the world I am proposing, there is a CI pass in that commits second parent (the tip being merged) and its first parent (previous tip of main) unless there was a flake and the previous tip of main failed (in which case it can/should be manually re-run, so, still). So I am proposing that every tag have *two* valid CI runs associated with it from the moment it is created, and a third with its first-parent (previous tip of main) ancestor.

> I have not changed anything.

As I said I don't know how/when the branch protection required checks were removed. My suggestion that it might have been incidental due to you renaming something was due to confusion on my part, I was thinking about your renaming of the docker workflows in the katzenqt repo.

> Between this merging and #1137 merging, a PR gets no docker-mixnet run at all.

This is our concern, dude.

> Note however that #1137 restores a per-PR docker run: its docker-interop-gate.yml triggers on pull_request and, when client code changes, runs the short lane (make dockertest_parallel=2 testargs=-short test, 20 min) plus the cross-version interop jobs; the full 2-hour suite stays on the schedule/tags. If you want a per-PR docker run continuously, merge #1137 first (it already carries the same docker-mixnet.yml trigger change) or merge #1171 and #1137 together.

I have other concerns with 1137 which we've discussed out of band, we can return to that discussion later but this is the first I've heard that the rationale for merging this rests on the assumption that that will land immediately after.

> As far as I understand the configuration the docker mixnet CI did not run on push to a branch.
[...]

> I think you are asking to _add a push trigger_ that is more than the pushing of a tag to trigger OR adding it as a required check. Both of which are fine with me but neither are something that I removed or changed. I removed only `on: [pull_request] ` in this pull request. You might be thinking about #1137, I suspect?

yes, this is a second point where i misspoke: i was confusing the `push` and `pull_request` triggers, sorry. it's `pull_request` we want. i agree we don't need docker ci on branches before a PR is opened.

> If you want to make it run once a week or once a month, I do not have a strong preference. I note that this enables the manual dispatch for the very first time.

infrequently makes even less sense to me: i am proposing that it always run on the latest main, and (either manually or with `pull_request`) on every PR. i see little benefit of running it a second time every once in a while, but sure it would waste less ci time if you made it less frequent.

anyway: I would like to figure out how to re-enable the required checks before we remove the `pull_request` trigger, because right now the `pull_request` trigger puts a red X there which discourages people from merging things that break it and I don't think that everyone will always manually run the check if it isn't required.

ioerror:
> > That doesn't test tags. On what basis do you object to a CI run of the docker mixnet for a specific tag when it is pushed? I would like tags to have success or failure directly associated with the tag being pushed. This is useful.
> 
> A tag is just associating a name with a commit. If CI has already passed on that commit, then there is a CI pass associated with the tag from the moment the tag is created. Also, in the world I am proposing, there is a CI pass in that commits second parent (the tip being merged) and its first parent (previous tip of main) unless there was a flake and the previous tip of main failed (in which case it can/should be manually re-run, so, still). So I am proposing that every tag have _two_ valid CI runs associated with it from the moment it is created, and a third with its first-parent (previous tip of main) ancestor.
> 

This policy proposal doesn't ensure every tag has a CI run that is derived from the tag being pushed.

As you note, perhaps (!) there is a flaky test, and perhaps it flaked twice or not at all. The tag run gives us something for the specific tagged release where we can check out the tag and begin to track it down. The change here ensures that pushing a tag gets the CI run. We can always do more, but do we already do what you suggest? If so, then the tag here is an extra CI run and indeed, three seems like two too many. Having tags tested is important to me because it says that when a tag fails, we shouldn't use that version unless we're certain that we understand the failures.

> > I have not changed anything.
> 
> As I said I don't know how/when the branch protection required checks were removed. My suggestion that it might have been incidental due to you renaming something was due to confusion on my part, I was thinking about your renaming of the docker workflows in the katzenqt repo.

Understood; the confusion is understandable as I recently moved a bunch of CI stuff in katzenqt to make it test against namenlos and more. It is finally running in under ~30 minutes now, so my hope was to take this CI from sometimes hours to far far less.

> 
> > Between this merging and #1137 merging, a PR gets no docker-mixnet run at all.
> 
> This is our concern, dude.

I understood and thus my suggestion was that one could 1) push the manual run after this was merged for the first time, 2) review and merge #1137 (which includes this change as well), or 3) for you to push a fixup commit that restores `on: [pull_request]` before merging it. I really don't feel strongly about removing that since #1137 essentially restores it.

> 
> > Note however that #1137 restores a per-PR docker run: its docker-interop-gate.yml triggers on pull_request and, when client code changes, runs the short lane (make dockertest_parallel=2 testargs=-short test, 20 min) plus the cross-version interop jobs; the full 2-hour suite stays on the schedule/tags. If you want a per-PR docker run continuously, merge #1137 first (it already carries the same docker-mixnet.yml trigger change) or merge #1171 and #1137 together.
> 
> I have other concerns with 1137 which we've discussed out of band, we can return to that discussion later but this is the first I've heard that the rationale for merging this rests on the assumption that that will land immediately after.
> 

#1137 contains this change but it has other changes included that create an end state that I think is mutually agreeable. However, I opened this PR because it relieves a pain point for people immediately and then the much shorter chained docker CI jobs  (and the multi hour CI jobs) are all settled in 1137.

> > As far as I understand the configuration the docker mixnet CI did not run on push to a branch.
> > [...]
> 
> > I think you are asking to _add a push trigger_ that is more than the pushing of a tag to trigger OR adding it as a required check. Both of which are fine with me but neither are something that I removed or changed. I removed only `on: [pull_request] ` in this pull request. You might be thinking about #1137, I suspect?
> 
> yes, this is a second point where i misspoke: i was confusing the `push` and `pull_request` triggers, sorry. it's `pull_request` we want. i agree we don't need docker ci on branches before a PR is opened.
> 

Okay, that makes much more sense to me, thank you for clarifying. 

Please feel free to push a commit on top that restores `on: [pull request]` if that is important to you for merging this request. If you however would prefer, we can close this request and move the review to #1137. No fixup commit is required there for this specific outcome, but perhaps other fixup commits are desired?

> > If you want to make it run once a week or once a month, I do not have a strong preference. I note that this enables the manual dispatch for the very first time.
> 
> infrequently makes even less sense to me: i am proposing that it always run on the latest main, and (either manually or with `pull_request`) on every PR. i see little benefit of running it a second time every once in a while, but sure it would waste less ci time if you made it less frequent.

My request was that if you see things about this that you would like to change, please make a concrete suggestion by pushing a fixup commit. If I agree, I will merge it, if not, I would return to discuss it further. I have not seen a concrete suggestion from you where I could 1) push a fixup commit to satisfy your pushback, or 2) review your fixup commit and merge it.

> anyway: I would like to figure out how to re-enable the required checks before we remove the `pull_request` trigger, because right now the `pull_request` trigger puts a red X there which discourages people from merging things that break it and I don't think that everyone will always manually run the check if it isn't required.

Are you asking me to make a specific change here? It seems like we've reached an impasse where one of us has to push something to satisfy the other person such that the other will merge the pull request. I touched the git tree last, so per my understanding of our (rather heavy handed) policy definitions, I am not allowed to merge without a positive review, and I do not see such a review. It seems that this pull request is effectively blocked for me and I am unable to satisfy the objections raised with any specific work on my end. Should I just close this pull request and we can continue on #1137?

